PR #1, opened 2019-11-27
feat: update to v19 params and new porep

dignifiedquire:
Needs https://github.com/filecoin-project/rust-fil-proofs/pull/958


dignifiedquire:
> wouldn't it make sense to also change in `rust-fil-proofs` the `window_size_nodes` from `usize` to `u32`?

it probably is

vmx:
Still approved (also with the new changes).

PR #15, opened 2019-12-09
feat: always rebuild local source 

vmx:
If `FFI_BUILD_FROM_SOURCE=1` is set, always rebuild from local source.

Change the scripts so that it is built within the `rust` directory
and not on a temporary directory that is different on every rebuild.
This way you get incremental builds if you make changes to the local
source (which is probably your intention if you build from source).

This PR also contains general improvements to the source (based on `shellcheck`
output).

### DIFF
--- a/install-shared.bash
+++ b/install-shared.bash
@@ -30,11 +30,10 @@ download_release_tarball() {
         --header "Accept:application/octet-stream" \
         --location \
         --output /dev/null \
-        -w %{url_effective} \
+        -w '%{url_effective}' \
         "{$__release_url}")
 
-    curl --retry 3 --output "${__tar_path}" "${__asset_url}"
-    if [[ $? -ne "0" ]]; then
+    if ! curl --retry 3 --output "${__tar_path}" "${__asset_url}"; then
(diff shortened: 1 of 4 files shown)


install-shared.bash
```diff
@@ -62,15 +60,8 @@ build_from_source() {
...
     cargo --version
-
-    if [[ -f "./scripts/build-release.sh" ]]; then
-        ./scripts/build-release.sh $__library_name $(cat rust-toolchain)
-    else
-        cargo build --release --all
-    fi
-
-    popd
+    cargo build --release --all
```
laser: Uh oh! We need to run the `build-release.sh` script in order to get the [proper linker flags to pack into the pkg-config file](https://github.com/filecoin-project/filecoin-ffi/blob/master/rust/scripts/build-release.sh#L31). 

Given that there's only one FFI lib now, artifacts like `install-shared.bash` and the `build-release.sh` can probably be collapsed into a single file (to make things a little more-obvious for maintainers).

vmx:
As things need to be built with the `build-release.sh` script, I'll do a new PR with a different approach.

In regards to the general changes, I've pushed them as a new PR: https://github.com/filecoin-project/filecoin-ffi/pull/16

PR #21, opened 2019-12-13
perf(proofs): parallelize replicas conversion in generate_candida…

dignifiedquire:
This helps quite considerably when passing large amounts of sectors `PrivateReplicaInfo::new()` does a disk read

### DIFF
--- a/rust/src/proofs/helpers.rs
+++ b/rust/src/proofs/helpers.rs
@@ -1,7 +1,7 @@
 use std::collections::btree_map::BTreeMap;
 use std::slice::from_raw_parts;
 
-use anyhow::Result;
+use anyhow::{Context, Result};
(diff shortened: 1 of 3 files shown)


rust/src/proofs/helpers.rs
```diff
@@ -135,33 +135,56 @@ pub fn bls_12_fr_into_bytes(fr: Fr) -> [u8; 32] {
...
-            })
-            .map(|p| (SectorId::from(ffi_r.sector_id), p))
+                comm_r,
+                replica_path,
+                sector_id,
+            } = info;
+
+            filecoin_proofs::PrivateReplicaInfo::new(replica_path, comm_r, cache_dir_path)
+                .map_err(|err| {
+                    format_err!(
```
vmx: Not sure if it's possible here (due to the iteration), but you might be able to use `.context()` for the error here.

PR #26, opened 2020-01-15
feat: implement explicit versioning for proofs

dignifiedquire:


laser:
@dignifiedquire - This is ready for review. I check out my edits to helpers.rs which accommodate a single replica winning multiple times.

laser:
There are many many methods defined in the file you linked to. Would you please tell me exactly which of the methods in that file need to be exposed through the FFI? I’d like to put this changeset to bed ASAP.

dignifiedquire:
for both post and porep we need

- `params_cid`
- `verifying_key_cid`
- `cache_params_path`
- `cache_verifying_key_path`
- `circuit_identifier`
- `version`

PR #27, opened 2020-01-15
feat(rust): update filecoin-proofs

dignifiedquire:
Depends on https://github.com/filecoin-project/rust-fil-proofs/pull/1017


whyrusleeping:
<3

PR #34, opened 2020-02-03
when region is China

jiaruihuang:
Obtaining filecoin proofs FFI release files is very slow when your region is China, and sometimes may fail. 
Each recompilation will download the release file, adding a judgment that if the file already exists, it does not need to be downloaded.


cryptonemo:
Closing due to inactivity

PR #37, opened 2020-02-18
Revert "fix: remove reference to .install-filecoin file (#20)"

vmx:
This reverts commit bbff56c621856352ac9d05540a32c5c48a2861fb.

This is a partial revert. It keeps the simplified instruction in the
README, but still does the Makefile tricks so that it works well with
parallel make. It also adds a comment about it, so that it doesn't
get removed again in the future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,9 @@ build tooling will attempt to compile a static library from local Rust sources.
 To opt out of downloading precompiled assets, set `FFI_BUILD_FROM_SOURCE=1`:
 
 ```shell
-FFI_BUILD_FROM_SOURCE=1 make clean all
+rm .install-filecoin \
+    ; make clean \
+    ; FFI_BUILD_FROM_SOURCE=1 make
 ```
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -18,7 +18,9 @@ build tooling will attempt to compile a static library from local Rust sources.
 To opt out of downloading precompiled assets, set `FFI_BUILD_FROM_SOURCE=1`:
 
 ```shell
-FFI_BUILD_FROM_SOURCE=1 make clean all
+rm .install-filecoin \
+    ; make clean \
```
Kubuxu: just `make clean` should be fine, right?

vmx: I thought :) but it isn't. It will clean, but say "nothing to be done" on the next run it will then actually do the compilation. I couldn't be bothered digging deeper :)

vmx:
@dignifiedquire any objections of merging this one?

dignifiedquire:
probably fine

PR #46, opened 2020-03-06
fix: attempt to fix verify-seal

dignifiedquire:


proofs.go
```diff
@@ -104,7 +104,7 @@ func VerifySeal(info abi.SealVerifyInfo) (bool, error) {
 	commRCBytes := C.CBytes(from32ByteArray(to32ByteArray(commR)))
 	defer C.free(commRCBytes)
 
-	proofCBytes := C.CBytes(info.OnChain.Proof)
+	proofCBytes := C.CBytes(info.OnChain.Proof[:])
```
laser: as per convo elsewhere: `Proof` is of type `[]byte`, so this slice is a noop

PR #47, opened 2020-03-11
generate CGO bindings with c-for-go

laser:
## Why does this PR exist?

Maintaining the CGO bindings by hand is error prone, challenging, and a bit of a waste of time (given that we can codegen most of it). The promise of this PR is that a developer will no longer need to write go-to-C (and back) mappers.

## What's in this PR?

This PR uses [c-for-go](https://github.com/xlab/c-for-go) to generate the bindings to libfilecoin.

## What's up with the huge diff?

There are a few things which contribute to the massively large diff-size:

1. c-for-go generates a Golang struct per C struct exposed through the FFI, as well as marshaling code. The combined proofs and BLS API is pretty large; there's a lot of code there.
1. By default, c-for-go does not generate bindings for your functions and structs. You need to give it a regex which it can use to built a whitelist of types for which it'll generate bindings. Prefixing all the FFI-exposed stuff with `fil_` made writing that regex quite easy (at the expense of a large diff).


README.md
```diff
@@ -21,6 +21,12 @@ To opt out of downloading precompiled assets, set `FFI_BUILD_FROM_SOURCE=1`:
 FFI_BUILD_FROM_SOURCE=1 make clean all
 ```
 
+## Updating CGO Bindings
+
+The CGO bindings are generated using [c-for-go](https://github.com/xlab/c-for-go).
+To generate bindings yourself, install the c-for-go binary, ensure that it's on
+your path, and then run `make cgo-gen`.
```
dignifiedquire: - Should have some notes about the fact that this is checked into source control. 
- Possible improvment,run the generation in CI and make sure the diff is zero?

Makefile
```diff
@@ -9,3 +9,7 @@ $(DEPS):
 clean:
 	rm -rf $(DEPS)
 .PHONY: clean
+
+cgo-gen:
+	c-for-go --ccincl --ccdefs --nostamp filecoin.yml
```
dignifiedquire: should we update the `clean` step to remove the generated files?

bls.go
```diff
@@ -13,22 +13,15 @@ import "C"
...
-	defer C.destroy_hash_response(resPtr)
-
-	// prep response
-	var digest Digest
-	digestSlice := C.GoBytes(unsafe.Pointer(&resPtr.digest), DigestBytes) // nolint: staticcheck
-	copy(digest[:], digestSlice)
-
-	return digest
+	resp := generated.FilHash(string(message), uint(len(message)))
+	resp.Deref()
```
dignifiedquire: what are the rules around these `Deref` calls?

Makefile
laser: Yeah, that's a good idea.

README.md
laser: > Possible improvment,run the generation in CI and make sure the diff is zero?

That's also a good idea. Thanks.

bls.go
laser: If Go receives a pointer from C, the caller must call `Deref()` on that pointer before its fields will be populated with a shallow copy of values from the C heap. `Deref()` will only unpack to a slice field if that slice has been initialized first, so sometimes two `Deref()` calls will be necessary. The `proofs.go` file should include sufficient examples such that further work in this area will be straightforward.

vmx:
Thanks @laser for making this happen. Code-gen those bindings makes so much sense.

PR #51, opened 2020-03-17
[WIP] feat: improve speed for default builds

dignifiedquire:
Depends on https://github.com/filecoin-project/rust-fil-proofs/pull/1060

Enables the following cpu feature flags for default builds

- Optional: `sha`, runtime detection will work around this missing


dignifiedquire:
Update: we are planning to do multiple prebuilts instead, to reduce the friction for folks.

dignifiedquire:
Resolved in #56

PR #55, opened 2020-03-18
feat: make logging to file possible

vmx:
The file is usually a pipe that was opened by the caller.

This feature was requested by @lanzafame (and earlier explored together with @Kubuxu iirc).

There are no Go binding yet, which I guess can be generated.


vmx:
This PR has a unit test and it shouldn't cause any harm as it just adds additional functionality. Even if it would break something, it would only be a problem when called. Hence I think we should merge this one and let Go implementers see if it works for them as expected.

vmx:
@dignifiedquire A review and potentially merge would be great. This repo had PRs in the past that never got merged, hence I'm pushing for it this time :)

PR #56, opened 2020-03-18
multiple released binaries

laser:
## Why does this PR exist?

We want to release a version of libfilecoin which takes advantage of some special instruction sets while simultaneously supporting CPUs which don't support those instruction sets.

## What's in this PR?

This changeset modifies our build pipeline to produce two versions of libfilecoin per operating system: standard (which is what we provide today) and optimized (which uses [dig's list of CPU features](https://filecoinproject.slack.com/archives/CP50PPW2X/p1584488024086800?thread_ts=1584474226.065700&cid=CP50PPW2X)).

This changeset also modifies the libfilecoin installation script to do some crude feature detection used to pick the correct version of libfilecoin to install (for your operating system and CPU).


install-filecoin
```diff
@@ -2,12 +2,137 @@
...
+
+    # determine where to look for CPU features
+    #
+    if [[ ! -f "/proc/cpuinfo" ]]; then
+        (>&2 echo "No /proc/cpuinfo file; falling back to Darwin feature detection")
+        searched="sysctl -a | grep machdep.cpu | tr '[:upper:]' '[:lower:]' | grep features"
+        features=$(eval "$searched")
+    else
+        features=$(cat /proc/cpuinfo | grep flags)
+    fi
```
vmx: Your code works, but I would write it as

```suggestion
    searched=""

    # determine where to look for CPU features
    #
    if [[ -f "/proc/cpuinfo" ]]; then
        searched="cat /proc/cpuinfo | grep flags"
    else
        (>&2 echo "No /proc/cpuinfo file; falling back to Darwin feature detection")
        searched="sysctl -a | grep machdep.cpu | tr '[:upper:]' '[:lower:]' | grep 
    fi
    features=$(eval "${searched}")
```

rust/scripts/build-release.sh
```diff
@@ -22,7 +22,10 @@ trap '{ rm -f $build_output_tmp; }' EXIT
 
 # build with RUSTFLAGS configured to output linker flags for native libs
 #
-RUSTFLAGS='--print native-static-libs' \
+rust_flags="${RUSTFLAGS}"
+rust_flags="--print native-static-libs ${rust_flags}"
```
vmx: Why not simply:

```suggestion
rust_flags="--print native-static-libs ${RUSTFLAGS}"
```

install-filecoin
laser: @vmx 

> Your code works, but I would write it as

good idea :)

install-filecoin
```diff
@@ -51,55 +51,55 @@ main() {
...
 
         if [[ ! -f "./libfilecoin.a" ]]; then
-            (>&2 echo "failed to install libfilecoin.a")
+            (>&2 echo "[install-filecoin/main] failed to install libfilecoin.a")
             exit 1
         fi
 
         if [[ ! -f "./filecoin.pc" ]]; then
-            (>&2 echo "failed to install filecoin.pc")
+            (>&2 echo "[main] failed to install filecoin.pc")
```
vmx:
```suggestion
            (>&2 echo "[install-filecoin/main] failed to install filecoin.pc")
```

rust/scripts/build-release.sh
laser: > Why not simply:

a fine idea!

install-filecoin
laser: done

.circleci/config.yml
```diff
@@ -208,6 +198,7 @@ workflows:
             branches:
               only:
                 - master
+                - feat/multiple-released-binaries
```
magik6k: Drop before merge

PR #62, opened 2020-03-24
Cosmetic changes

vmx:
I thought now is a good time to beautify the code base a bit. It's a pure refactor without any functiona changed.

### DIFF
--- a/rust/src/lib.rs
+++ b/rust/src/lib.rs
@@ -1,12 +1,6 @@
 #![deny(clippy::all)]
 #![allow(clippy::missing_safety_doc)]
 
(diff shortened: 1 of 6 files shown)


rust/src/lib.rs
```diff
@@ -1,12 +1,6 @@
-#![deny(clippy::all, clippy::perf, clippy::correctness)]
+#![deny(clippy::all)]
```
dignifiedquire: when did this change? when I added these `clippy::all` did not contain these

vmx: No idea, it's been like this in "my forever".

PR #63, opened 2020-03-24
gitingore filecoin.[h, pc]

magik6k:


laser:
Hm. filecoin.h and filecoin.pc are old build assets; as per jbenet, the assets were renamed to filcrypto.h and filcrypto.pc on Monday.

PR #70, opened 2020-04-06
test: run logging with pipe test on Linux only

vmx:
OSX also doesn't support `libc::pipe2`, hence run the test on Linux only.


vmx:
For future reference, if you want cross platform pipes, use https://crates.io/crates/os_pipe.

PR #73, opened 2020-04-09
regenerate CGO bindings + drop unused struct which referenced nonexistent abi type

laser:
## Things in PR

This changeset fixes the CGO compilation errors (mismatches between generated CGO and the header file) and the non-CGO errors, too.

This changeset also replaces the now-antiquated usage of paramcache with go-paramfetch, much like storage-fsm and the like.

It also updates some Rust issues that clippy caught.


parameters.json
```diff
@@ -0,0 +1,122 @@
+{
```
dignifiedquire, line 1: why did you add these into this code base?

laser: @dignifiedquire 

> why did you add these into this code base?

CircleCI uses this file (see the .circleci/config.yaml in this PR) to download Groth parameters. Previously, we were `cargo install`'ing paramcache and generating them ourselves based on a rust-fil-proofs Git SHA, which was a nightmare.

PR #79, opened 2020-04-27
fix: error if there is already a logger

vmx:
The file logger needs to be initialized as the first action when calling the
FFI. If there was already any logger initialized, return an error when trying
to initialize the file logger.


vmx:
Ping, this is more or less a trivial one, that starts to get stale :)

vmx:
I merge this one as it should only improve the situation and not make anything worse.

PR #80, opened 2020-04-27
use dtynn's forked c-for-go to regenerate CGO bindings

laser:
## Why does this PR exist?

The c-for-go binary generates code that, under certain conditions, produces ENOMEM errors. A very helpful user, @dtynn, has patched c-for-go - but [their patch has yet to be merged into upstream](https://github.com/xlab/c-for-go/pull/91).

## What's in this PR?

This PR updates CGO bindings to those generated by dtynn's patch. It also updates CircleCI to generate using that same, forked c-for-go.

## Warnings

With the last lotus network reset, we are no longer able to reproduce this issue. So, we don't know if this patch actually fixes the issues. We're going to merge, deploy, and if it causes problems - we'll revert and pursue some other fix.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -294,9 +294,16 @@ commands:
   ensure_generated_cgo_up_to_date:
     steps:
       - run:
-          name: Generate CGO bindings and compare with what's tracked by Git
+          name: Generate CGO bindings (using forked c-for-go) and compare with what's tracked by Git
           command: |
-            go run github.com/xlab/c-for-go --ccincl --ccdefs --nostamp filcrypto.yml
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -296,8 +296,10 @@ commands:
       - run:
           name: Generate CGO bindings and compare with what's tracked by Git
           command: |
-            go run github.com/xlab/c-for-go --ccincl --ccdefs --nostamp filcrypto.yml
```
dignifiedquire, line 299: why not check using the fork?

laser: fixed

PR #95, opened 2020-05-20
add changelog and changelog generator

laser:
## Why does this PR exist?

We need to get in the habit of combining a release with a changelog update so that we can easily communicate the features and bugfixes that go out with each release.

## What's in this changeset?

This changeset includes a changelog starting at v0.25.0 to current. It also includes a changelog-generating script which is [based off of the tool which go-ipfs uses](https://github.com/ipfs/go-ipfs/blob/master/bin/mkreleaselog).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,98 @@
+# filecoin-ffi changelog
+
+## 0.26.2
+
+This release contains a fix for a bug which prevented unmodified miners from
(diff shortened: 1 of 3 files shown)


vmx:
I'd propose we switch to [Conventional Commits](https://www.conventionalcommits.org/) (most of us use those already anyway). This way we could use existing tooling to create changelogs and wouldn't need our own custom tooling.

PR #111, opened 2020-06-23
Update master dependencies

cryptonemo:
fix: remove some cbindgen warnings
feat: update some dependencies
feat: add build.sh and run_tests.sh convenience scripts


build.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+set -e
+
+make clean
+cd rust
+rm Cargo.lock
+cargo update -p "filecoin-proofs-api"
+cargo install cbindgen
```
laser, line 9: @cryptonemo - Concern: I don't think that the `cargo install cbindgen` command [will respect](https://github.com/rust-lang/cargo/issues/7169) the version of cbindgen specified in our our Cargo.lock file, which means that `cargo build` (which uses `build.rs`) could produce different output than the output produced by the `cbindgen` binary. In rare cases, this may result in CircleCI builds performing differently than local builds.

run_tests.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+RUST_LOG=info go test -count=1 ./...
```
laser, line 3: Suggestion: Run the Rust tests here as well.

rust/Cargo.toml
```diff
@@ -26,15 +26,19 @@ rand = "0.7"
...
+bellperson = { version = "0.9.0", features = ["gpu"] }
 serde_json = "1.0.46"
 
 [dependencies.filecoin-proofs-api]
-package = "filecoin-proofs-api"
-version = "4.0.0"
+git = "https://github.com/filecoin-project/rust-filecoin-proofs-api"
+branch = "master"
+
+#[dependencies.filecoin-proofs-api]
```
laser, line 36: Question: What was the rationale for pulling this stuff in from Git instead of crates.io?

go.mod
```diff
@@ -10,4 +10,12 @@ require (
 	github.com/libp2p/go-libp2p-core v0.3.0 // indirect
 	github.com/pkg/errors v0.9.1
 	github.com/stretchr/testify v1.4.0
+	github.com/tj/go-spin v1.1.0 // indirect
```
laser, line 13: @cryptonemo - Suggestion: It looks like this stuff was added erroneously. If that's true, run `go mod tidy` to clean this (and `go.sum`) up.

rust/src/proofs/api.rs
```diff
@@ -1009,7 +1009,7 @@ pub unsafe extern "C" fn fil_get_post_version(
     registered_post_proof_accessor(registered_proof, |p| Ok(format!("{:?}", p)))
 }
 
-unsafe fn registered_seal_proof_accessor(
+pub unsafe fn registered_seal_proof_accessor(
```
laser, line 1012: Question: What is the intent of making these public? Do you wish to expose them to Go?

build.sh
cryptonemo: Ah, interesting.  I added it to solve the reverse problem (Cargo.lock specified a version too old to succeed for me, so I updated Cargo.toml).  If I just remove `cargo install cbindgen`, it should still work, no?

rust/Cargo.toml
cryptonemo: I just wanted to point master to `-api` master for now.  That will go away when we have a new point release.

rust/src/proofs/api.rs
cryptonemo: There were warnings about them, no other intent.

build.sh
cryptonemo: Actually, this script isn't for CI, it's for the end user, so I don't think it should matter (other than perhaps not be needed).  I needed it for some reason though.

rust/Cargo.toml
cryptonemo: Back story -- right this moment, -api has some perf changes in master that points to proofs master, and we want some testing done that needs ffi to be buildable, so this is a way to go about it.

cryptonemo:
Feedback changes coming soon ...

run_tests.sh
cryptonemo: Updated.  But bad news, rust tests are failing.

cryptonemo: Some of which I think are just missing params.  But some of which are not.  Any ideas @laser ?

laser: > Any ideas @laser ?

Not offhand, no. I'll check the branch out and give it a test.

PR #116, opened 2020-06-29
handle null pointer returned from fil_aggregate

laser:
## Why does this PR exist?

An error from fil_aggregate is encoded as a NUL byte. Unlike other FFI "Response" structs, there is no error code and message. As such, we need to check for the Go nil value before dereferencing the inner signature (which produced a nil-pointer exception today).

### DIFF
--- a/bls.go
+++ b/bls.go
@@ -61,7 +61,9 @@ func HashVerify(signature *Signature, messages []Message, publicKeys []PublicKey
 	return isValid > 0
 }
 
-// Aggregate aggregates signatures together into a new signature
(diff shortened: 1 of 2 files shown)


bls.go
```diff
@@ -61,7 +61,9 @@ func HashVerify(signature *Signature, messages []Message, publicKeys []PublicKey
 	return isValid > 0
 }
 
-// Aggregate aggregates signatures together into a new signature
+// Aggregate aggregates signatures together into a new signature. If the
```
cryptonemo, line 64: Something seems off about this comment

PR #118, opened 2020-07-01
add FauxRep to ffi package

laser:


porcuquine:
@cryptonemo Please hold off on releasing this until I resolve a mystery. I'm seeing some errors I shouldn't be elsewhere and want to make sure there's not something off in the dependency chain this will unleash.

porcuquine:
I have isolated the problem I was observing, and I no longer think it is related to anything happening here. You can now safely ignore my previous comment.

PR #123, opened 2020-07-22
fix: check for empty message/signature before verify

cryptonemo:


cryptonemo:
This PR explicitly fails on empty incoming parameters.  It previously failed in any case.  This PR may need to be adjusted as preferred usage may in fact be to succeed in this case.

cryptonemo:
Perhaps not needed

PR #128, opened 2020-07-30
fix: pull in fix-darwin neptune branch to resolve issue

cryptonemo:


cryptonemo:
Note that this PR will not be merged, but it does show working CI with a neptune patch for darwin.  When the patch release is available, we can issue a new `ffi` PR that simply uses the latest published crate version, rather than the pinned 'older' version it uses on `master` now.

ribasushi:
@cryptonemo is this still applicable?

cryptonemo:
> @cryptonemo is this still applicable?

It's no longer necessary now that the issue is resolved.  I don't think we're in a rush to release the updated neptune version, as I believe it's slated for NSE related releases.

PR #133, opened 2020-09-03
Consume go-state-types

ZenGround0:
This is blocked on specs actors consuming go-state-types on release branch 9 at upcoming tag v0.9.4 

FYI @anorth @Stebalien


proofs.go
```diff
@@ -131,7 +132,7 @@ func VerifyWindowPoSt(info abi.WindowPoStVerifyInfo) (bool, error) {
 
 // GeneratePieceCommitment produces a piece commitment for the provided data
 // stored at a given path.
-func GeneratePieceCID(proofType abi.RegisteredSealProof, piecePath string, pieceSize abi.UnpaddedPieceSize) (cid.Cid, error) {
+func GeneratePieceCID(proofType aabi.RegisteredSealProof, piecePath string, pieceSize abi.UnpaddedPieceSize) (cid.Cid, error) {
```
Stebalien, line 135:
```suggestion
func GeneratePieceCID(proofType aabi.RegisteredSealProof, piecePath string, pieceSize aabi.UnpaddedPieceSize) (cid.Cid, error) {
```

ZenGround0: ah these should both be `abi`

anorth:
I recommend waiting for https://github.com/filecoin-project/specs-actors/pull/1096

anorth:
These `SealVerifyInfo` and friends are currently in specs-actors, not go-state-types, so there will be two versions of them. Presumably we don't want to version this ffi package with release branches, though (it will need to add new functions if things change). We'll alias on to the types from the 0.9 branch. Until we do so, though, we might need  a release branch for testnet.

anorth:
https://github.com/filecoin-project/specs-actors/pull/1097

Stebalien (approved):
LGTM but I have _no_ idea how we'd switch between proof types at runtime. For the next few network upgrades, we'll probably be fine here, but this could be a bit tricky.

ZenGround0:
Yes we should definitely put this on the list of things we need to figure out for future upgrades.

PR #134, opened 2020-09-08
feat: return faulty sectors

vmx:
If Window PoSt fails, return the faulty sectors. They are part of the
error response. The `faulty_sectors_len` is the number of faulty sectors
(0 if there aren't any faulty sectors) and `faulty_sectors_ptr` points
to an array of unsigned 64-bit integers, representing the sector IDs.

### DIFF
--- a/proofs.go
+++ b/proofs.go
@@ -17,6 +17,7 @@ import (
 	"github.com/filecoin-project/specs-actors/actors/runtime/proof"
 	"github.com/ipfs/go-cid"
 	"github.com/pkg/errors"
+	"golang.org/x/xerrors"
 
 	"github.com/filecoin-project/filecoin-ffi/generated"
 )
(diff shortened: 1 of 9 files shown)


rust/src/proofs/api.rs
```diff
@@ -1722,4 +1735,169 @@ pub mod tests {
...
+                randomness,
+                private_replicas.as_ptr(),
+                private_replicas.len(),
+                prover_id,
+            );
+
+            assert_eq!(
+                (*resp_j).status_code,
+                FCPResponseStatus::FCPUnclassifiedError,
+                "generate_winning_post should have failed"
```
porcuquine, line 1881: 'generate_window_post', I think.

porcuquine (approved):

vmx:
This should *NOT* be mereged until the corresponding Go changes (which should be pretty minimal) are in.

magik6k (approved):

proofs.go
```diff
@@ -808,6 +808,14 @@ func fromFilPoStFaultySectors(ptr []uint64, l uint) ([]abi.SectorNumber, error)
 		return nil, nil
 	}
 
+	type sliceHeader struct {
+		Data unsafe.Pointer
+		Len  int
+		Cap  int
+	}
+
+	(*sliceHeader)(unsafe.Pointer(&ptr)).Len = int(l) // don't worry about it
```
Kubuxu, line 817: Is this needed? Isn't just the fact that we are inerating up to `l` below enough?

magik6k: The generated cgo code doesn't set it, si we panic on out-of-bounds

whyrusleeping (approved):

PR #136, opened 2020-09-22
feat: point to blst branch for the deadlock fix

cryptonemo:


cryptonemo:
No longer needed (new `fil-blst` crate was released)

PR #146, opened 2020-10-13
fix: update proofs release to latest

cryptonemo:

PR #147, opened 2020-10-14
fix: manually install hwloc and statically link against it

cryptonemo:


cryptonemo:
No longer needed -- important things were added to #146

PR #150, opened 2020-11-02
feat: proofs upgrade for sdr expander truncation

cryptonemo:
feat: test legacy and non-legacy registered seal proof types

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -386,11 +386,11 @@ commands:
             - run:
                 name: Run leak detector
                 command: make cgo-leakdetect
-                no_output_timeout: 60m
+                no_output_timeout: 90m
       - run:
           name: Run the Rust tests
           command: cd rust && FIL_PROOFS_PARAMETER_CACHE="${HOME}/filecoin-proof-parameters/" RUST_LOG=info cargo test --all --release && cd ..
-          no_output_timeout: 60m
+          no_output_timeout: 90m
       - run:
           name: Run the Go tests
           command: GODEBUG=cgocheck=2 RUST_LOG=info go test -p 1 -timeout 60m
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -389,7 +389,7 @@ commands:
                 no_output_timeout: 90m
       - run:
           name: Run the Rust tests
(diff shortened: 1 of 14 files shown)


porcuquine:
A datum: `test_get_gpu_devices` takes 0 secs for me on two different machines (one with and one without usable GPUs).

porcuquine:
Sealing tests do seem to hang when there is no GPU, though.

porcuquine:
I was able to reproduce the hanging behavior (on a machine without GPU). I can also seemingly 'fix' it by deleting orphaned lock files. The possibility of such is not likely fixable on the timeframe of this issue, but let me see if I can come up with a bandaid for the symptoms.

porcuquine:
>Maybe related -- @magik6k reported that the test_get_gpu_devices test is taking up to 30 seconds. I am currently running it and it's taking 10 seconds. I'm not sure what the normal amount of time is at the moment though. The sealing tests do complete properly for me when run in parallel w/o issue in ~1.5 minutes.

I think this is not related. I believe it is caused by the switch to use gpu tools, which tries to look up all AMD or Nvidia devices and potentially introduces a ten second pause if the corresponding platform does not exist. (@cryptonemo This is what was happening repeatedly, leading to very slow CI times. That doesn't happen any more, but the pause on first access is 'unavoidable' as the code is written currently.)

I don't believe this is the cause of the hanging. That seems to be caused by an (inexplicable so far) inability to get an exclusive file lock on the priority lock file the first time it is accessed.

cryptonemo:
Something to consider: I do most of my dev/testing on a machine w/o a GPU and the tests complete properly (i.e. no hangs like CI is showing).

dignifiedquire:
I have reproduced them on my Mac and Porcu was able to as well. They only happen when the `gpu` features are enabled and no gpus are available I believe. Trying some things atm here https://github.com/filecoin-project/bellperson/tree/fix-locking/src

dignifiedquire:
I have switched to using only 1 thread for the tests, avoiding this race condition for the moment in CI. The original issue has been around for a while, but was triggered due to the additional tests that this PR adds. Fixing them will likely be more involved as it seems.

magik6k (approved):

PR #154, opened 2020-11-12
Feat/cgo improvements

xlab:
A few important improvements related to major pain-points of the generated bindings.

Previously, the bindings contained workarounds:
* Custom allocators for C structs because CGO didn't like non-copying pointers to Go's heap memory (resolved with https://github.com/xlab/c-for-go/commit/ea6dce5809cbc4c80e20e3673ca783979a9b425a)
* A lot of `string`<>`[]byte` typecast with risk of getting a buffer to be cut because of `\x00` (resolved with https://github.com/xlab/c-for-go/commit/582a2294915dfba95c2e89dd070d4c999dd22ebc)

After resolving the core issues in c-for-go toolchain, I've regenerated bindings and adjusted the code so it could take benefit of these new features.

Basically, the changes are:
* Always copy byte slices using `C.CByte` and allocate
* Always copy strings using `C.CString` and allocate
* Don't treat `const uint8_t *` a.k.a. `const unsigned char *` as a string. Only `const char *` counts by default. That distinction helped to isolate strings (i.e. error messages, paths, etc) from the buffers of data.

All tests ran with `./run-tests.sh` pass OK.


cryptonemo:
Try running `make cgo-gen` and then committing and pushing those changes.

generated/cgo_helpers.go
```diff
@@ -242,11 +242,17 @@ func allocFilClearCacheResponseMemory(n int) unsafe.Pointer {
 
 const sizeOfFilClearCacheResponseValue = unsafe.Sizeof([1]C.fil_ClearCacheResponse{})
 
-// unpackPCharString represents the data from Go string as *C.char and avoids copying.
+// unpackPCharString copies the data from Go string as *C.char.
 func unpackPCharString(str string) (*C.char, *cgoAllocMap) {
+	allocs := new(cgoAllocMap)
+	defer runtime.SetFinalizer(allocs, func(a *cgoAllocMap) {
```
Kubuxu, line 248: Unrelated but something I came up with:
If you had function `GoFree` on the cgoAllocMap (and different alloc maps) that has the content of this lambda.
Then it can be shortened to:
`runtime.SetFinalizer(allocs, (*cgoAllocMap).GoFree)` while eliminating the need for a closure.

Kubuxu: See: https://play.golang.org/p/55AytJ-XjXz

Kubuxu (approved):
Looks good to me

Kubuxu:
I'll merge it latter today unless someone raises something. cc @magik6k

generated/cgo_helpers.go
xlab: @Kubuxu that's a cool shortcut!

PR #155, opened 2020-12-01
feat: update dependencies, including latest proofs

cryptonemo:
fix: add back in rust native target building

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -293,7 +293,7 @@ commands:
             cd rust
 
             TARBALL_PATH="/tmp/${CIRCLE_PROJECT_REPONAME}-$(uname)-optimized.tar.gz"
-            RUSTFLAGS="-C target-feature=$(cat rustc-target-features-optimized.json | jq -r '.[].rustc_target_feature' | tr '\n' ',')"
+            RUSTFLAGS="-C target-cpu=native -C target-feature=$(cat rustc-target-features-optimized.json | jq -r '.[].rustc_target_feature' | tr '\n' ',')"
 
             ./scripts/build-release.sh filcrypto $(cat ./rust-toolchain) --verbose --locked --no-default-features --features pairing
             ./scripts/package-release.sh $TARBALL_PATH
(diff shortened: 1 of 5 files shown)


dignifiedquire:
@Kubuxu @magik6k  important note: using this in lotus, requires to upgrade blst to `0.3.2`

.circleci/config.yml
```diff
@@ -293,7 +293,7 @@ commands:
             cd rust
 
             TARBALL_PATH="/tmp/${CIRCLE_PROJECT_REPONAME}-$(uname)-optimized.tar.gz"
-            RUSTFLAGS="-C target-feature=$(cat rustc-target-features-optimized.json | jq -r '.[].rustc_target_feature' | tr '\n' ',')"
+            RUSTFLAGS="-C target-cpu=native -C target-feature=$(cat rustc-target-features-optimized.json | jq -r '.[].rustc_target_feature' | tr '\n' ',')"
```
dignifiedquire, line 296: @cryptonemo I don't think this is a good idea, this will specialize to the CPU that circle ci has, ending up in randomly hard to detect issues. If folks want to build against native, they will need to build for themselves.

dignifiedquire (changes requested):

.circleci/config.yml
cryptonemo: These builds aren't published. Maybe better to remove the CI jobs?

dignifiedquire: ah right, I always get confused with this CI job

rust/Cargo.lock
```diff
@@ -28,9 +28,9 @@ dependencies = [
 
 [[package]]
```
dignifiedquire, line 29: @cryptonemo this file should not be removed, we need to have a fixed Cargo.lock

cryptonemo: I'm not following.  Cargo.lock was updated to the latest deps, not removed.

dignifiedquire: damn, I am not on my game today, there where so many deletions and got confused

dignifiedquire (approved):

cryptonemo:
@dignifiedquire Thanks for the review, but was hoping @magik6k or @Kubuxu could also approve if it looks good/tests well.

dignifiedquire:
Do not merge this PR yet, the blst update is likely to break in a way need to first figure out to work around

cryptonemo:
> Do not merge this PR yet, the blst update is likely to break in a way need to first figure out to work around

Right.  It won't be merged until all is resolved there.

cryptonemo:
No longer needed

PR #157, opened 2020-12-10
Add Version const

Kubuxu:
To avoid issues similar to the most recent upgrade where,
interface did not change but new proof type got introduced and some
users did not upgrade filecoin-ffi in Lotus.


cryptonemo (approved):
Looks good. This is a version only used in the go code, correct?

Kubuxu:
Yeah.

PR #161, opened 2021-01-27
feat: add blst-portable option for compatibility

cryptonemo:
This PR adds a blst portable option by adding a build flag `FFI_USE_BLST_PORTABLE=1`.

This PR also uses the portable option when building and publishing the standard blst pre-built binary (for maximum compatibility).

Requires https://github.com/filecoin-project/blstrs/pull/17 and https://github.com/filecoin-project/bls-signatures/pull/34

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -273,7 +273,8 @@ commands:
             TARBALL_PATH="/tmp/${CIRCLE_PROJECT_REPONAME}-$(uname)-standard-pairing.tar.gz"
             RELEASE_NAME="${CIRCLE_PROJECT_REPONAME}-$(uname)-standard-pairing"
 
-            ./scripts/build-release.sh filcrypto $(cat ./rust-toolchain) --verbose --locked --no-default-features --features pairing
+            # Note: the blst dependency uses the portable configuration for maximum compatibility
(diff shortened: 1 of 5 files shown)


dignifiedquire:
please document the flag in the readme

.circleci/config.yml
```diff
@@ -284,7 +284,8 @@ commands:
             TARBALL_PATH="/tmp/${CIRCLE_PROJECT_REPONAME}-$(uname)-standard-blst.tar.gz"
             RELEASE_NAME="${CIRCLE_PROJECT_REPONAME}-$(uname)-standard-blst"
 
-            ./scripts/build-release.sh filcrypto $(cat ./rust-toolchain) --verbose --locked --no-default-features --features blst
+            # Note: the standard blst release uses the portable configuration for maximum compatibility
+            ./scripts/build-release.sh filcrypto $(cat ./rust-toolchain) --verbose --locked --no-default-features --features blst --features blst-portable
```
dignifiedquire, line 288: if we want to change the default, we need to do so for the pairing one too, as it uses `blst` now for signatures as well

cryptonemo: Ah, didn't realize, thanks!

Kubuxu (approved):

PR #162, opened 2021-01-27
feat: re-factor unseal_range to use get_unsealed_range for performance

cryptonemo:
Requires https://github.com/filecoin-project/rust-fil-proofs/pull/1401


cryptonemo:
This PR doesn't yet do anything (other than re-factores code) until the dependency code merges.  Once the other code merges, this should be able to be updated with the latest dependencies to pick it up automatically.

rust/src/proofs/api.rs
```diff
@@ -421,16 +421,17 @@ pub unsafe extern "C" fn fil_unseal_range(
...
+        let sealed_sector = std::fs::File::from_raw_fd(sealed_sector_fd_raw);
+        let unseal_output = std::fs::File::from_raw_fd(unseal_output_fd_raw);
 
-        let result = filecoin_proofs_api::seal::unseal_range(
+        let result = filecoin_proofs_api::seal::get_unsealed_range(
             registered_proof.into(),
             c_str_to_pbuf(cache_dir_path),
-            &mut sealed_sector,
-            &mut unseal_output,
+            sealed_sector.path().unwrap(),
```
dignifiedquire, line 433: these might be pipes, I don' tthink there is a gurantee that there will always be a filepath available.

cryptonemo: Yes, I'm aware.  Unfortunately, a caller re-factor or audit is also required to make use of this.

magik6k: `sealed_sector` is currently passed as a normal file

cryptonemo: Great, thanks!  What about the unseal output?  Right now I am also assuming it's a normal file, but there are options to work around that if we have to.

rust/src/proofs/api.rs
```diff
@@ -421,16 +421,17 @@ pub unsafe extern "C" fn fil_unseal_range(
...
+            unseal_output.path().unwrap(),
```
magik6k, line 434: unseal_output is always an anonymous pipe -> https://github.com/filecoin-project/lotus/blob/master/extern/sector-storage/ffiwrapper/sealer_cgo.go#L323

We do that because lotus-miner wants to have the same data format for sectors passed into PreCommit, and unsealed data, so unsealed sector files look like
```
[32GiB of fr32 padded data][RLE+ of allocated byte ranges][4B int of that RLE+ length]
```

cryptonemo: Ok, I can probably leave that argument alone then.  Thanks

magik6k:
Also potentially worth noting - lotus will currently always unseal the whole sector as it seems that there really isn't any advantage to not doing that - https://github.com/filecoin-project/lotus/blob/master/extern/sector-storage/manager.go#L297-L300

rust/src/proofs/api.rs
cryptonemo: An alternative is to just have a new API call, or a versioned call for unseal_range or something

cryptonemo:
> Also potentially worth noting - lotus will currently always unseal the whole sector as it seems that there really isn't any advantage to not doing that - https://github.com/filecoin-project/lotus/blob/master/extern/sector-storage/manager.go#L297-L300

Unfortunately, that's the case for us as well.  Even though we have ranges, they are subsets after a full sector unseal.

rust/src/proofs/api.rs
cryptonemo: It might be easiest to have a new call with slightly different args, and I think we can hide this new call at the FFI layer, so I don't think anything would change from the caller side.  I'll see if there's a way around that, but not sure at the moment.

cryptonemo: Basically, I want this to be public, so we can use that: https://github.com/filecoin-project/rust-fil-proofs/blob/fix-ram-unseal/filecoin-proofs/src/api/mod.rs#L261

cryptonemo: Or some small wrapper around it, like the other unseal calls now are

cryptonemo: Hah, nevermind, we already have what I want (as of this PR) -- just need to plumb it through: https://github.com/filecoin-project/rust-fil-proofs/blob/fix-ram-unseal/filecoin-proofs/src/api/mod.rs#L195

cryptonemo:
Fixed the broken test case!

cryptonemo:
Rebased against master

cryptonemo:
Resolves https://github.com/filecoin-project/rust-fil-proofs/issues/1402

cryptonemo:
Requires a few things:

1) A new versioned proofs release,
2) An updated version of this PR that points to that version: https://github.com/filecoin-project/rust-filecoin-proofs-api/pull/56/files,
3) An updated version of *this* PR that has updated dependencies

Once approved, I'll get the other items in place before this is merged.
cc: @magik6k

rust/src/proofs/api.rs
magik6k: Unseal output is always an anonymous pipe - on the Go side it's just piped into a fr32 padded file with a trailer containing an rle which marks which parts of the file are allocated

See https://github.com/filecoin-project/lotus/blob/b54b1303f9f6bcc3b57ae25cfc1174b59f44b943/extern/sector-storage/ffiwrapper/sealer_cgo.go#L322-L362

magik6k (commented):
Looks good, guessing that we need to fix the pipe support

cryptonemo:
I left the unseal output alone -- only the sealed sector will need to be used as a file for this to work (because we retrieve the path when calling `get_unsealed_range_mapped` from the fd, which won't work on the pipe of course)

magik6k (approved):

PR #163, opened 2021-01-30
feat: add -Zfeatures=itarget to support conditional compilation of dependency feature based on target

dgbo:
This is related to https://github.com/filecoin-project/rust-fil-proofs/pull/1404.
We want to enable the feature `compress, asm` on aarch64 for `sha2`, while only enable feature `compress` on `x86`.
The `-Zfeatures=itarget` is necessary for supporting conditional compilation of dependency feature based on target [1][2].

[1] https://github.com/rust-lang/cargo/issues/2524
[2] https://github.com/rust-lang/cargo/issues/7914

### DIFF
--- a/install-filcrypto
+++ b/install-filcrypto
@@ -150,12 +150,13 @@ build_from_source() {
     cargo --version
 
     # Add feature specific rust flags as needed here.
+    additional_flags="-Zfeatures=itarget"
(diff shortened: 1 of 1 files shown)


dgbo:
With https://github.com/RustCrypto/hashes/pull/224 merged into sha-0.9.3, this is not needed anymore.

vmx:
I think this is fixed meanwhile, feel free to re-open if it's not.

PR #166, opened 2021-02-24
feat: update dependencies to use proof aggregation branches

cryptonemo:
This is a WIP getting the dependencies required for proof aggregation hooked in. The API is not wired through yet, but will be added here.


cryptonemo:
The API has been wired in now and end to end aggregation testing works via the `--ignored` rust tests.

rust/src/proofs/api.rs
```diff
@@ -402,6 +404,166 @@ pub unsafe extern "C" fn fil_seal_commit_phase2(
...
+            }
+        }
+
+        info!("seal_commit_phase2_for_aggregation: finish");
+
+        raw_ptr(response)
+    })
+}
+
+pub unsafe extern "C" fn fil_aggregate_seal_proofs(
```
magik6k, line 473: It seems like the Go method for this didn't get generated

cryptonemo: Ok, I've got most of the re-factor working locally, but am not sure how it affects your commit.  I will likely open a new draft PR on a different branch and we can resolve which is better later.

cryptonemo:
This PR is outdated now compared to this alternative: https://github.com/filecoin-project/filecoin-ffi/pull/169

PR #169, opened 2021-03-10
Add proof aggregation API

cryptonemo:
This PR is building on https://github.com/filecoin-project/filecoin-ffi/pull/166, so it's an alternative to it.


proofs.go
```diff
@@ -57,9 +57,63 @@ func VerifySeal(info proof.SealVerifyInfo) (bool, error) {
 	return resp.IsValid, nil
 }
 
+func VerifyAggregateSeals(aggregate proof5.AggregateSealVerifyProofAndInfos) (bool, error) {
+	if len(aggregate.Infos) == 0 {
+		return false, xerrors.New("no seal verify infos")
+	}
+
+	spt := aggregate.Infos[0].SealProof // todo assuming this needs to be the same for all sectors, potentially makes sense to put in AggregateSealVerifyProofAndInfos
```
anorth, line 65: FYI @ZenGround0 

@cryptonemo do you know if this assumption is true?

cryptonemo: The assumption is true, but it's not technically constrained.  I believe we can aggregate different proof types, but I think for now we are just choosing not to do that.  The FIP should be updated that they should all be the same though since that's the intention for now.

rust/src/proofs/types.rs
```diff
@@ -230,6 +253,37 @@ impl Drop for fil_VanillaProof {
...
+            status_code: FCPResponseStatus::FCPNoError,
+            error_msg: ptr::null(),
+            proof_len: 0,
+            proof_ptr: ptr::null(),
+        }
+    }
+}
+
+impl Drop for fil_AggregateProof {
+    fn drop(&mut self) {
```
dignifiedquire, line 276: I think this leaks the `error_msg` if there is one

cryptonemo: It's not allocated/managed, it just points to a rust string from the ffi toolkit as far as I know (similar to all other usages of the `error_msg`)

vmx: If you implement `Drop` manually, you also need to drop `error_msg`. In other structs the `#[derive(DropStructMacro)]` implements `Drop`. Though in this case you can just use the `DropStructMacro` derive instead (see https://github.com/filecoin-project/rust-fil-ffi-toolkit/tree/master/drop-struct-macro-derive#readme for more information).

cryptonemo: Thanks, I see, will take a closer look.  At one point, I may have needed to impl Drop rather than use the macro, but not sure that's the case any more.

magik6k (approved):
Go side looks good

cryptonemo:
Requires https://github.com/filecoin-project/rust-fil-proofs/pull/1395 and https://github.com/filecoin-project/rust-filecoin-proofs-api/pull/51 and updated proofs releases before merging.

magik6k (approved):

PR #177, opened 2021-05-03
fix: use correct default features for the published binaries

cryptonemo:


cryptonemo:
@travisperson The darwin build environment looks broken -- let me know if you can approve anyway or have ideas on how to resolve that.

travisperson (approved):

cryptonemo:
Hrm, I just realized OS X is required, so I can't merge anyway right now.

cryptonemo:
@dignifiedquire Do you have git powers to merge this PR?

dignifiedquire:
afaik we also generate prebuilts on macos, so we should fix those builds before triggering a new set of prebuilts

cryptonemo:
> afaik we also generate prebuilts on macos, so we should fix those builds before triggering a new set of prebuilts

Hrm, yes we do -- didn't realize.  I've asked Infra to take a look at the CI machine.

PR #178, opened 2021-05-06
fix: update deps to proofs v7.0.1

cryptonemo:


ribasushi (approved):
Tested on nerpa - builds and seals CC

jennijuju (approved):

PR #179, opened 2021-05-27
Build universal binaries for darwin

hannahhoward:
# Goals

fix https://github.com/filecoin-project/lotus/issues/5697 or at least start to fix it

# Implementation

- per last comment I am able to compile Lotus on Apple M1 but only using FFI_BUILD_FROM_SOURCE
- I believe this is cause the prebuilts are generated from the CircleCI release pipeline, which runs XCode 10.0.0. XCode 10.0.0 will generate binaries that are arch x86/amd64 specific. M1 Support was added in XCode 12.
- update xcode used in circleci to generate prebuilt binaries. this should mean the binary built is a
univseral binary 
- note there may be more work to do -- this is a bit of a guess

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -318,16 +318,10 @@ commands:
             TARBALL_PATH="/tmp/${CIRCLE_PROJECT_REPONAME}-$(uname)-aarch64-standard-pairing.tar.gz"
             RELEASE_NAME="${CIRCLE_PROJECT_REPONAME}-$(uname)-aarch64-standard-pairing"
 
-            # Temporarily move the rust-toolchain file since that won't allow it to be used
-            #mv ./rust-toolchain ./rust-toolchain.bak
-            #rustup default nightly-aarch64-apple-darwin
-
             # Note: the blst dependency uses the portable configuration for maximum compatibility
             ./scripts/build-release.sh filcrypto $(cat rust-toolchain) --target aarch64-apple-darwin --verbose --locked --no-default-features --features multicore-sdr --features pairing,gpu --features blst-portable
             ./scripts/package-release.sh $TARBALL_PATH
             ./scripts/publish-release.sh $TARBALL_PATH $RELEASE_NAME
-
-            #mv ./rust-toolchain.bak ./rust-toolchain
       - run:
           name: Build and publish the standard release for x86_64 (blst)
           command: |
@@ -349,9 +343,8 @@ commands:
             TARBALL_PATH="/tmp/${CIRCLE_PROJECT_REPONAME}-$(uname)-aarch64-standard-blst.tar.gz"
             RELEASE_NAME="${CIRCLE_PROJECT_REPONAME}-$(uname)-aarch64-standard-blst"
 
-            rustup default nightly-aarch64-apple-darwin
             # Note: the blst dependency uses the portable configuration for maximum compatibility
-            ./scripts/build-release.sh filcrypto nightly-aarch64-apple-darwin --verbose --locked --no-default-features --features multicore-sdr --features blst,gpu --features blst-portable
+            ./scripts/build-release.sh filcrypto $(cat rust-toolchain) --target aarch64-apple-darwin --verbose --locked --no-default-features --features multicore-sdr --features blst,gpu --features blst-portable
             ./scripts/package-release.sh $TARBALL_PATH
             ./scripts/publish-release.sh $TARBALL_PATH $RELEASE_NAME
   configure_environment_variables:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -104,6 +104,8 @@ jobs:
           linux: false
           darwin: true
       - run: cd rust && rustup install $(cat rust-toolchain)
+      - run: cd rust && rustup install nightly-aarch64-apple-darwin
+      - run: cd rust && rustup target add aarch64-apple-darwin
       - run: cd rust && rustup default $(cat rust-toolchain)
       - run: cd rust && cargo fetch
       - publish_darwin_release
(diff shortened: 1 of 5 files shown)


cryptonemo:
Hrm, I don't know one way or the other if this will work.  Could be nice if it does, because I don't want to add another pre-built for OS X if it can be avoided.

This should be an issue related/linked to one where `lotus` pre-built binaries should be considered, rather than `ffi` pre-built binaries.

hannahhoward:
Just to be clear, this adds no new binaries-- there is still only a single Darwin binary. It just makes the Darwin binaries compatible with both x86 and M1

momack2:
@cryptonemo - can we move this forward?

cryptonemo:
> @cryptonemo - can we move this forward?

If it's just a matter of updating the xcode version on CI and someone is able to verify this does the right thing, yes.  @hannahhoward have you verified the binary works in both cases?

vmx:
Slightly related, we really *should* update to a new Xcode version on CI, as the one we currently use (10.0.0) [will be deprecated in September](https://discuss.circleci.com/t/xcode-deprecation-notice-9-4-1-10-0-0-10-1-0-and-10-2-1/40515?mkt_tok=NDg1LVpNSC02MjYAAAF-AoXdcDb_dzg0mn2Cb6Qlj2acScsSZ23L540HHs1cgTH5I_X3n5QSvIH_2vX9MRsw06gC8ySrv6JWVCUXVTxd-90hDY0aN9Ysf98CEYdzfEM).

momack2:
@cryptonemo - can you own landing (and testing) this? Feel free to draft others via slack to test on darwin hardware if needed!

cryptonemo:
I agree with @vmx that we should be updating Xcode, but noting here that the binary produced as it is now fails on M1, complaining that it was built for `x86_64`.

cryptonemo:
I'll try a few more things to see if this can happen.  Basing some testing on some notes here:

https://github.com/rust-lang/cargo/issues/8875
https://github.com/rust-lang/cargo/issues/2024
https://developer.apple.com/documentation/apple-silicon/building-a-universal-macos-binary

cryptonemo:
It doesn't look like our CI will allow us to cross compile to aarch64.  For a universal binary, it looks like you need to build both `x86_64` and `aarch64` binaries and then you roll them together after the fact to create it (using `lipo`).  At the moment, I haven't figured out a way to get CI to build the aarch64 binary, so this is stalled unless someone can assist there.

```
info: using existing install for 'nightly-aarch64-apple-darwin'
info: default toolchain set to 'nightly-aarch64-apple-darwin'

  nightly-aarch64-apple-darwin unchanged - (error reading rustc version)

+ main filcrypto nightly-aarch64-apple-darwin --verbose --locked --no-default-features --features multicore-sdr --features pairing,gpu --features blst-portable
+ [[ -z filcrypto ]]
+ [[ -z nightly-aarch64-apple-darwin ]]
++ mktemp
+ __build_output_log_tmp=/var/folders/bl/wbxjgtzx7j5_mjsmfr3ynlc00000gp/T/tmp.YSkW53Ak
+ trap '{ rm -f $__build_output_log_tmp; }' EXIT
+ local '__rust_flags=--print native-static-libs '
+ RUSTFLAGS='--print native-static-libs '
+ cargo +nightly-aarch64-apple-darwin build --release --verbose --locked --no-default-features --features multicore-sdr --features pairing,gpu --features blst-portable
+ tee /var/folders/bl/wbxjgtzx7j5_mjsmfr3ynlc00000gp/T/tmp.YSkW53Ak
error: command failed: 'cargo'

Caused by:
    Bad CPU type in executable (os error 86)
+ rm -f /var/folders/bl/wbxjgtzx7j5_mjsmfr3ynlc00000gp/T/tmp.YSkW53Ak

Exited with code exit status 1

CircleCI received exit code 1
```

cryptonemo:
Some more relevant info: https://support.circleci.com/hc/en-us/articles/360056461452-Apple-M1-Apple-Silion-Support-on-CircleCI

https://circleci.canny.io/cloud-feature-requests/p/support-new-m1-arm-based-macs

cryptonemo:
Got something to build, but have to verify tomorrow that it's a proper aarch64 binary (rather than a native one).  If it worked, I can continue looking into using lipo to make a universal.

cryptonemo:
> Got something to build, but have to verify tomorrow that it's a proper aarch64 binary (rather than a native one). If it worked, I can continue looking into using lipo to make a universal.

Just to follow-up here -- this is the same issue as previously found.  Specifying the build target doesn't seem to work and emits an x86_64 binary, so it's not compatible.  This cross compiling approach on CI does not appear to be working.

momack2:
> XCode 10.0.0 will generate binaries that are arch x86/amd64 specific. M1 Support was added in XCode 12.

Are you saying this analysis by @hannahhoward is incorrect?

cryptonemo:
> > XCode 10.0.0 will generate binaries that are arch x86/amd64 specific. M1 Support was added in XCode 12.
> 
> Are you saying this analysis by @hannahhoward is incorrect?

Not necessarily.

The work in this PR shows that the original PR submission does not work and the premise that simply upgrading to XCode 12 will generate universal binaries is incorrect.  I found that several days ago (and anyone can download the published binaries attached to the commits of this PR and verify).

There are notes I linked above that say the way to do this is to:

1) Generate an x86_64 binary,
2) Generate an aarch64 binary,
3) Use a tool called `lipo' to roll them together into a universal binary

At that point we can test if it's the real deal or not.

All of the above can be done with XCode 12, according to the Apple docs.

This PR is stalled because I can't find a way to cross compile the OS X binary from x86_64 to aarch64 in the CI environment (requirement 2).  Until that's resolved, requirement 3 can't happen.  I also linked some notes stating the challenges of trying to do that on our CI environment specifically (i.e. it isn't official supported in CI yet, but might be at some point).

dignifiedquire (approved):

PR #186, opened 2021-06-16
don't require rustup

coryschwartz:
rustup isn't actually required to build.

The specific reason I'm requesting this change that the macos brew installation could include rust as a dependency, but it seems inappropriate to run rustup-init as part of the build. Instead, I should be able to use the system rust toolchain.

There are probably other situations where you want to build FFI and use the system rust.

### DIFF
--- a/install-filcrypto
+++ b/install-filcrypto
@@ -139,12 +139,6 @@ build_from_source() {
         exit 1
     fi
 
-    if ! [ -x "$(command -v rustup)" ]; then
-        (>&2 echo '[build_from_source] Error: rustup is not installed.')
-        (>&2 echo '[build_from_source] install Rust toolchain installer to resolve this problem.')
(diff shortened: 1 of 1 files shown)


cryptonemo:
Should probably check for `cargo` and `rust` though, no?

coryschwartz:
oh sure, I can change the check for 'rust' rather than 'rustup'

dignifiedquire:
If you do this, you should check that `rustc --version` matches whatever is in https://github.com/filecoin-project/filecoin-ffi/blob/master/rust/rust-toolchain otherwise you will break the assumptions made. Rustup will use the version specified in that file, but if you use a system rust install that will not be the case, and things might start to break suddenly

vmx:
I'm closing this PR as using rustup is the preferred way to do it.

PR #197, opened 2021-08-20
* init decoupling-post-worker code

gegezai:
Merge based on：
https://github.com/filecoin-project/lotus/discussions/6943

### DIFF
--- a/types.go
+++ b/types.go
@@ -2,6 +2,7 @@ package ffi
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"sort"
 
@@ -89,14 +90,20 @@ func (s *SortedPublicSectorInfo) UnmarshalJSON(b []byte) error {
 
 // NewSortedPrivateSectorInfo returns a SortedPrivateSectorInfo
 func NewSortedPrivateSectorInfo(sectorInfo ...PrivateSectorInfo) SortedPrivateSectorInfo {
-	fn := func(i, j int) bool {
-		return bytes.Compare(sectorInfo[i].SealedCID.Bytes(), sectorInfo[j].SealedCID.Bytes()) == -1
+	result := make([]PrivateSectorInfo, 0)
+	seen := map[abi.SectorNumber]struct{}{}
+	for i := range sectorInfo {
+		if _, found := seen[sectorInfo[i].SectorNumber]; !found {
+			seen[sectorInfo[i].SectorNumber] = struct{}{}
+			result = append(result, sectorInfo[i])
+		}
 	}
-
-	sort.Slice(sectorInfo[:], fn)
+	sort.Slice(result, func(i, j int) bool {
+		return result[i].SectorNumber < result[j].SectorNumber
+	})
 
 	return SortedPrivateSectorInfo{
-		f: sectorInfo,
+		f: result,
 	}
 }
 
@@ -130,4 +137,12 @@ type PrivateSectorInfo struct {
 // AllocationManager is an interface that provides Free() capability.
 type AllocationManager interface {
 	Free()
+}
+
+func SplitSortedPrivateSectorInfo(ctx context.Context, sortPrivSectors SortedPrivateSectorInfo, start int, end int) (SortedPrivateSectorInfo, error) {
+	var newSortPrivSectors SortedPrivateSectorInfo
+	newSortPrivSectors.f = make([]PrivateSectorInfo, 0)
+	newSortPrivSectors.f = append(newSortPrivSectors.f, sortPrivSectors.f[start:end]...)
+
+	return newSortPrivSectors, nil
 }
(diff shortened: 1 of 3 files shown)


gegezai:
@jennijuju The code has been debugged and verified, please review the merge

types.go
```diff
@@ -131,3 +132,50 @@ type PrivateSectorInfo struct {
 type AllocationManager interface {
 	Free()
 }
+
+func DeDuplicatePrivateSectorInfo(ctx context.Context, sortPrivSectors *SortedPrivateSectorInfo) (SortedPrivateSectorInfo, error) {
```
Kubuxu, line 136: Looking at code in Lotus it looks like it should be possible to execute this and split operations before creating the `SortedPrivateSectorInfo`.
I strongly suggest changing it to that.
The SortedPrivateSectorInfo type should only be used for directly passing sector information into FFI not for carrying sector information around or operating on them.

Kubuxu (changes requested):
As requested in the comment, perform operations required first and then create an instance of SortedPrivateSectorInfo.

Changes in proofs.go and distributed.go look good.

gegezai:
@Kubuxu Agree with you. We have a try

gegezai:
@jennijuju  @Kubuxu According to the review comments, the deduplication of SortedPrivateSectorInfo is performed in lotus, because the key member variables of SortedPrivateSectorInfo are private types and cannot be accessed at the lotus layer. 
image

We tried to perform deduplication before calling ffi.NewSortedPrivateSectorInfo and tested it, but proof verification Fail.

Any suggestions for modification?

types.go
```diff
@@ -131,3 +132,50 @@ type PrivateSectorInfo struct {
...
+		flag := false
+		for j := 0; j < new_sector_len-i-1; j++ {
+			if newSortPrivSectors.f[j].SectorNumber > newSortPrivSectors.f[j+1].SectorNumber {
+				newSortPrivSectors.f[j], newSortPrivSectors.f[j+1] = newSortPrivSectors.f[j+1], newSortPrivSectors.f[j]
+				flag = true
+			}
+		}
+		if !flag {
+			break
+		}
```
Kubuxu, line 160: The above code seems to sort sector infos by SectorNumber, but the original code sorts them by SealedCID, why the change?

Also if you want to sort, please use `sort.Slice`. See types.go#93

gegezai: @Kubuxu According to the logic before splitting,Split calculation results need to be sorted by SectorNumber
企业微信截图_16377349048939

gegezai: > 

@Kubuxu Any suggestions for modification?

gegezai: > 

Because the members of the SortedProvateSectorInfo structure are private and cannot be operated in lotus, we think the DeDuplicatePrivateSectorInfo interface is necessary, or are there any other implementation suggestions? @Kubuxu

magik6k: Won't passing sectors deduplicated here to `NewSortedPrivateSectorInfo` sort them correctly? If not, can we change `NewSortedPrivateSectorInfo` to sort correctly?

gegezai: @magik6k We can change NewSortedPrivateSectorInfo to sort correctly, If It without affecting the logic of other functions

gegezai: @magik6k Modification has been submitted, please review

types.go
```diff
@@ -89,14 +90,37 @@ func (s *SortedPublicSectorInfo) UnmarshalJSON(b []byte) error {
 
 // NewSortedPrivateSectorInfo returns a SortedPrivateSectorInfo
 func NewSortedPrivateSectorInfo(sectorInfo ...PrivateSectorInfo) SortedPrivateSectorInfo {
-	fn := func(i, j int) bool {
-		return bytes.Compare(sectorInfo[i].SealedCID.Bytes(), sectorInfo[j].SealedCID.Bytes()) == -1
+	var remove_duplicate_privSector = make([]PrivateSectorInfo, 0)
```
magik6k, line 93: Maybe use a shorter variable name? We're using it in almost every line of this function. (we can also preallocate this slice)

```suggestion
	result := make([]PrivateSectorInfo, 0, len(sectorInfo))
```

types.go
```diff
@@ -89,14 +90,37 @@ func (s *SortedPublicSectorInfo) UnmarshalJSON(b []byte) error {
...
+		}
 	}
 
-	sort.Slice(sectorInfo[:], fn)
+	new_sector_len := len(remove_duplicate_privSector)
+	if new_sector_len < 2 {
+		return SortedPrivateSectorInfo{
+			f: remove_duplicate_privSector,
+		}
+	}
```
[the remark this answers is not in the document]
gegezai: Accept

gegezai:
@magik6k  The code has been debugged and verified, please review and merge

magik6k (approved):
I still don't really understand the change in sorting - why sorting by SealedCID(CommR) worked before, and why we're changing it to sort by SectorNumber. Reading the rust code, the only thing I see is that the code doesn't really seem to care about sorting, at least in case of the monolith post methods.

Will merge this, if it's somehow broken it should be really easy to see that and fix

PR #198, opened 2021-08-24
fix: use released crates

vmx:
There is no longer the need to use Git based dependencies, the fixes
that were needed for proper compilation on MacOS were released.

This is the proper fix for the MacOS linker issue. I've only updated the minimum of what's needed in the Cargo.lock. I'd like to get this one merged, even if it's too late for the current release, to make sure it's definitely in the next one.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -356,8 +356,9 @@ dependencies = [
 
 [[package]]
 name = "cl3"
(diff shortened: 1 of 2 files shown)


cryptonemo (approved):

vmx:
@cryptonemo I leave it for you to merge as you better know when it is a good time to do so.

PR #205, opened 2021-10-01
feat: add cuda to CI (didn't realize it was supported)

cryptonemo:
fix: add blstrs gpu features

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -51,6 +51,6 @@ tempfile = "3.0.8"
 [features]
 default = ["opencl", "multicore-sdr" ]
 blst-portable = ["bls-signatures/blst-portable", "blstrs/portable"]
-opencl = ["filecoin-proofs-api/opencl", "bellperson/opencl", "storage-proofs-porep/opencl", "rust-gpu-tools/opencl"]
-cuda = ["filecoin-proofs-api/cuda", "bellperson/cuda", "storage-proofs-porep/cuda", "rust-gpu-tools/cuda"]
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -251,6 +251,16 @@ commands:
 
             ./scripts/build-release.sh filcrypto $(cat ./rust-toolchain) build --verbose --locked --no-default-features --features multicore-sdr,opencl
             ./scripts/package-release.sh $TARBALL_PATH
+      - run:
+          name: Build the optimized release (cuda)
```
dignifiedquire, line 255: do we not run tests on these?

cryptonemo: Apparently not, only default features seem to have the tests run

cryptonemo: I'll remove those builds and swap for tests

cryptonemo: Failed with no nvcc, as I originally expected. Can try to add cuda via apt I guess

cryptonemo: No nvidia-cuda-toolkit.  What am I missing?

cryptonemo: Ah, I see, the ubuntu image

cryptonemo: No go yet, I've removed everything except for the blstrs update for now.  I'll create an issue for CI testing of CUDA but don't want it to block the release

vmx: On CI you need to make sure that the CUDA stuff (`/usr/local/cuda-11.2/bin`) is in the `PATH`, see e.g. https://github.com/filecoin-project/rust-gpu-tools/blob/d3b36477c7d2017e2b7ea284fd91342a667835db/.circleci/config.yml#L26

vmx:
These days CUDA is enabled by default, hence running on CI.

PR #215, opened 2021-11-17
Use the new ReplicaUpdateInfo type from actors

arajasek:
Use the new `ReplicaUpdateInfo` type from actors


Kubuxu:
The `FunctionsSectorUpdate` is just a syntax wrapper.
Verification is for example meant to be used as: `ffi.SectorUpdate.VerifyUpdate`

We can switch to the existing way (throwing everything into proofs.go).

PR #217, opened 2021-12-17
FVM integration

raulk:


rust/src/fvm/machine.rs
```diff
@@ -0,0 +1,264 @@
...
+    let next_id = NEXT_ID.fetch_add(1, std::sync::atomic::Ordering::SeqCst);
+    let mut machines = FVM_MAP.lock().unwrap();
+    machines.insert(next_id, machine);
+    next_id
+}
+
+fn get_default_config() -> fvm::Config {
+    Config {
+        initial_pages: 1024, //FIXME
+        max_pages: 32768,    // FIXME
```
raulk, line 50: @Stebalien any sense of the values we want to set here? Noting that WASM page size is 64KiB, so we may want to start with 16 pages (1MiB), and have no upper limit for M0? We probaly want to instrument the WASM runtime to record the pages we ended up allocating with every execution.

rust/src/fvm/machine.rs
```diff
@@ -0,0 +1,264 @@
...
+#[cfg(not(target_os = "windows"))]
+pub unsafe extern "C" fn fil_create_fvm_machine(
+    fvm_version: fil_FvmRegisteredVersion,
+    chain_epoch: u64,
+    token_amount: u64,
+    network_version: u64,
+    state_root_ptr: *const u8,
+    state_root_len: libc::size_t,
+    blockstore_id: u64,
+    externs_id: u64,
```
raulk, line 69: I _think_ externs can be stateless if the blockstore is _not_ folded into the externs, but need to think more about this.

rust/src/fvm/machine.rs
```diff
@@ -0,0 +1,264 @@
...
+/// Note: the incoming args as u64 and odd conversions to i32/i64
+/// for some types is due to the generated bindings not liking the
+/// 32bit types as incoming args
+///
+#[no_mangle]
+#[cfg(not(target_os = "windows"))]
+pub unsafe extern "C" fn fil_create_fvm_machine(
+    fvm_version: fil_FvmRegisteredVersion,
+    chain_epoch: u64,
+    token_amount: u64,
```
raulk, line 64: Is this the basefee? Token amounts are BigInts, [effectively bounded at 91 bits](https://github.com/filecoin-project/fvm/issues/56) given the current maximum supply in AttoFiL (unit of exchange).

rust/src/fvm/machine.rs
```diff
@@ -0,0 +1,264 @@
...
+    catch_panic_response(|| {
+        init_log();
+
+        info!("fil_create_fvm_machine: start");
+
+        let mut response = fil_CreateFvmMachineResponse::default();
+
+        let config = get_default_config();
+        let chain_epoch = chain_epoch as ChainEpoch;
+        let token_amount = TokenAmount::from_u64(token_amount);
```
raulk, line 80: Can't be u64, needs to be u128 if safe for FFI, or a tuple of (hi, lo) u64s.

rust/src/fvm/machine.rs
```diff
@@ -0,0 +1,264 @@
...
+                    Ok(x) => x,
+                    Err(err) => {
+                        response.status_code = FCPResponseStatus::FCPUnclassifiedError;
+                        response.error_msg = rust_str_to_c_str(format!("{:?}", err));
+                        return raw_ptr(response);
+                    }
+                };
+
+                response.status_code = FCPResponseStatus::FCPNoError;
+                // FIXME: Return relevant fields of ApplyRet
```
raulk, line 206: Can return ApplyRet as (DAG-)CBOR encoded.

rust/src/fvm/machine.rs
```diff
@@ -0,0 +1,264 @@
...
+            todo!("execute message")
+        }
+        None => {
+            todo!("invalid machine id")
+        }
+    }
+
+    info!("fil_fvm_machine_flush_message: end");
+    // })
+}
```
raulk, line 243: I don't think we have a "finish message" operation. What we finish is the _machine_, which flushes all effectively linked blocks to the blockstore, and returns the new state root CID.

@Stebalien do we expect the node to construct the receipts root, or do we want to accumulate receipts in the machine, to then return the root?

rust/src/fvm/types.rs
```diff
@@ -0,0 +1,135 @@
...
+    pub to_len: libc::size_t,
+    pub sequence: u64,
+    pub value: u64,
+    pub method_num: u64,
+    pub params_bytes: *const u8,
+    pub params_len: libc::size_t,
+    pub gas_limit: u64,
+    pub gas_fee_cap: u64,
+    pub gas_premium: u64,
+}
```
raulk, line 64: We can use a DAG-CBOR-encoded message (which is how it's transmitted in the network anyway), and deserialize it in the Rust-side FFI handler, instead of modelling it again.

rust/src/fvm/machine.rs
Stebalien: No upper limit for now. We probably also need an exception for cron.

And yeah, we need to measure it. Then 10x at least.

Stebalien: Note: I'd actually base the max memory on the total callstack. So we'll configure some max pages here, then set max_pages to `max_pages - used_pages` when instantiating a new instance.

rust/src/fvm/machine.rs
Stebalien: We usually just use bigints.

rust/src/fvm/machine.rs
Stebalien: I expect the node to deal with that, honestly. Basically:

1. The VM handles the state-tree _only_.
2. The Node handles everything related to the blockchain.

cryptonemo:
Currently requires https://github.com/filecoin-project/ref-fvm/pull/277

rust/src/proofs/types.rs
```diff
@@ -862,7 +862,7 @@ impl Clone for fil_SealCommitPhase2Response {
 code_and_message_impl!(fil_SealCommitPhase2Response);
 
 #[repr(C)]
-#[derive(Clone, DropStructMacro)]
+#[derive(Clone, Default, DropStructMacro)]
```
cryptonemo, line 865: Good catch, thanks

rust/src/fvm/types.rs
```diff
@@ -0,0 +1,96 @@
...
+
+code_and_message_impl!(fil_CreateFvmMachineResponse);
+
+#[repr(C)]
+#[derive(DropStructMacro)]
+pub struct fil_FvmMachineExecuteResponse {
+    pub error_msg: *const libc::c_char,
+    pub status_code: FCPResponseStatus,
+    pub exit_code: u64,
+    pub return_ptr: *const u8,
```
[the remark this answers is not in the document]
vmx: I think it's correct, it follows the usual pattern.

.circleci/config.yml
```diff
@@ -7,7 +7,7 @@ orbs:
 executors:
   golang:
     docker:
-      - image: circleci/golang:1.13
+      - image: circleci/golang:1.16.4
```
mvdan, line 10: FYI this version is significantly out of date in terms of bugfix releases, the latest is 1.16.15. 1.16.x is also unsupported now that 1.18 is out, but I imagine you're already aware of that.

cgo/blockstore.go
```diff
@@ -0,0 +1,130 @@
...
+	"unsafe"
+
+	blocks "github.com/ipfs/go-block-format"
+	"github.com/ipfs/go-cid"
+	blockstore "github.com/ipfs/go-ipfs-blockstore"
+)
+
+/*
+#include <stdint.h>
+typedef const uint8_t* buf_t;
```
dignifiedquire, line 13: Why the use of `buf_t`? seems like this typedef should rather just live in go land if you want to use it

cgo/extern.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+}
+
+//export cgo_extern_verify_consensus_fault
+func cgo_extern_verify_consensus_fault(
+	handle C.uint64_t,
+	h1 C.buf_t, h1Len C.int32_t,
+	h2 C.buf_t, h2Len C.int32_t,
+	extra C.buf_t, extraLen C.int32_t,
+	minerId *C.uint64_t,
```
dignifiedquire, line 69: would be more readable to prefix these out variables with sth like `out_`

cgo/registry.go
```diff
@@ -0,0 +1,49 @@
+package cgo
+
+import (
+	"context"
+	"sync"
+)
+
+var (
+	mu       sync.RWMutex
```
dignifiedquire, line 9: @Stebalien are you planning on pulling the atomic version into this?

fvm.go
```diff
@@ -0,0 +1,242 @@
...
+	executor unsafe.Pointer
+}
+
+const (
+	applyExplicit = iota
+	applyImplicit
+)
+
+// CreateFVM
+func CreateFVM(fvmVersion uint64, externs cgo.Externs, epoch abi.ChainEpoch, baseFee abi.TokenAmount, baseCircSupply abi.TokenAmount, nv network.Version, stateBase cid.Cid) (*FVM, error) {
```
dignifiedquire, line 38: can we maybe use a config struct here, these are a lot of args

fvm.go
```diff
@@ -0,0 +1,242 @@
...
+	baseFeeHi, baseFeeLo, err := splitBigInt(baseFee)
+	if err != nil {
+		return nil, xerrors.Errorf("invalid basefee: %w", err)
+	}
+	baseCircSupplyHi, baseCircSupplyLo, err := splitBigInt(baseCircSupply)
+	if err != nil {
+		return nil, xerrors.Errorf("invalid circ supply: %w", err)
+	}
+
+	exHandle := cgo.Register(context.TODO(), externs)
```
dignifiedquire, line 48: what's the plan regarding contexts here?

fvm.go
```diff
@@ -0,0 +1,242 @@
...
+		f.executor = nil
+		generated.FilDropFvmMachine(executor)
+		cgo.Unregister(exHandle)
+	})
+
+	return fvm, nil
+}
+
+func (f *FVM) ApplyMessage(msgBytes []byte, chainLen uint) (*ApplyRet, error) {
+	defer runtime.KeepAlive(f)
```
dignifiedquire, line 87: why is this needed?

fvm.go
```diff
@@ -0,0 +1,242 @@
...
+}
+
+// returns hi, lo
+func splitBigInt(i big.Int) (hi uint64, lo uint64, err error) {
+	if i.Sign() < 0 {
+		return 0, 0, xerrors.Errorf("negative number: %s", i)
+	}
+	words := i.Bits()
+	switch bits.UintSize {
+	case 32:
```
dignifiedquire, line 166: filecoin-ffi until now does not support 32bit systems, do we need to do so now?

rust/Cargo.toml
```diff
@@ -38,22 +38,36 @@ memmap = "0.7"
...
 
 [features]
 default = ["opencl", "multicore-sdr" ]
 blst-portable = ["bls-signatures/blst-portable", "blstrs/portable"]
-opencl = ["filecoin-proofs-api/opencl", "bellperson/opencl", "storage-proofs-porep/opencl", "rust-gpu-tools/opencl"]
-cuda = ["filecoin-proofs-api/cuda", "bellperson/cuda", "storage-proofs-porep/cuda", "rust-gpu-tools/cuda"]
+opencl = ["filecoin-proofs-api/opencl", "bellperson/opencl", "storage-proofs-porep/opencl", "rust-gpu-tools/opencl", "fvm/opencl"]
+cuda = ["filecoin-proofs-api/cuda", "bellperson/cuda", "storage-proofs-porep/cuda", "rust-gpu-tools/cuda", "fvm/cuda"]
 multicore-sdr = ["storage-proofs-porep/multicore-sdr"]
+bindgen=[]
```
dignifiedquire, line 73: what's this feature for?

rust/src/fvm/blockstore/cgo.rs
```diff
@@ -0,0 +1,139 @@
...
+
+use anyhow::{anyhow, Result};
+use cid::Cid;
+use fvm_shared::blockstore::Blockstore;
+
+use super::super::cgo::*;
+
+const ERR_NO_STORE: i32 = -1;
+const ERR_NOT_FOUND: i32 = -2;
+const MAX_BUF_SIZE: usize = 4 << 20; // 4MiB
```
dignifiedquire, line 11: can we get some comments as to what these constants exactly specify?

rust/src/fvm/blockstore/cgo.rs
```diff
@@ -0,0 +1,139 @@
...
+                    e => Err(anyhow!("cgo blockstore 'put' failed with error code {}", e)),
+                }
+            }
+        }
+
+        let mut lengths = Vec::with_capacity(MAX_BLOCK_BATCH);
+        let mut buf = Vec::with_capacity(MAX_BUF_SIZE);
+        for (k, block) in blocks {
+            let block = block.as_ref();
+            if lengths.len() >= MAX_BLOCK_BATCH
```
dignifiedquire, line 105: why is this needed?

rust/src/fvm/blockstore/fake.rs
```diff
@@ -0,0 +1,99 @@
...
+    fn get(&self, k: &Cid) -> Result<Option<Vec<u8>>> {
+        match self.fake_blocks.borrow().get(k) {
+            Some(blk) => Ok(Some(blk.clone())),
+            None => self.base.get(k),
+        }
+    }
+
+    fn put_keyed(&self, k: &Cid, block: &[u8]) -> Result<()> {
+        // TODO: we won't need this hack once we merge
+        // https://github.com/filecoin-project/ref-fvm/pull/382.
```
dignifiedquire, line 39: this is merged, can we remove it?

rust/src/fvm/blockstore/overlay.rs
```diff
@@ -0,0 +1,69 @@
+use std::collections::HashMap;
+
+use anyhow::Result;
+use cid::Cid;
+use fvm_shared::blockstore::Blockstore;
+
+pub struct OverlayBlockstore<BS> {
```
dignifiedquire, line 7: please add some docs on what the purpose of this blockstore is

rust/src/fvm/externs.rs
```diff
@@ -0,0 +1,128 @@
...
+                e => Err(anyhow!(
+                    "cgo extern 'verify_consensus_fault' failed with error code {}",
+                    e
+                )),
+            }
+        }
+    }
+}
+
+impl Externs for CgoExterns {}
```
dignifiedquire, line 128: what's the purpose of this trait?

rust/src/fvm/machine.rs
```diff
@@ -0,0 +1,276 @@
...
+    DefaultKernel<DefaultCallManager<DefaultMachine<OverlayBlockstore<CgoBlockstore>, CgoExterns>>>,
+>;
+
+lazy_static! {
+    static ref ENGINE: fvm::machine::Engine = fvm::machine::Engine::default();
+}
+
+/// Note: the incoming args as u64 and odd conversions to i32/i64
+/// for some types is due to the generated bindings not liking the
+/// 32bit types as incoming args
```
dignifiedquire, line 32: This is very suspicious, and easy to get wrong. What were the issues you had here?

rust/src/fvm/machine.rs
```diff
@@ -0,0 +1,276 @@
...
+        // Only do this if the return data is non-empty. The empty vec pointer is non-null and not
+        // valid in go.
+        if !apply_ret.msg_receipt.return_data.is_empty() {
+            let return_bytes = Vec::from(apply_ret.msg_receipt.return_data).into_boxed_slice();
+            response.return_ptr = return_bytes.as_ptr();
+            response.return_len = return_bytes.len();
+            Box::leak(return_bytes);
+        }
+
+        // TODO: Do something with the backtrace.
```
dignifiedquire, line 197: is this expected to be done before being merged?

rust/src/fvm/machine.rs
```diff
@@ -0,0 +1,276 @@
...
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn fil_drop_fvm_machine(executor: *mut libc::c_void) {
+    let _ = Box::from_raw(executor as *mut Mutex<CgoExecutor>);
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn fil_fvm_machine_execute_message(
+    executor: *mut libc::c_void,
```
dignifiedquire, line 143: why is this a void pointer, instead of a pointer to a concrete type?

cgo/blockstore.go
Stebalien: IIRC, I had issues with the "const" part. Although maybe we don't need that? This came from the "just work dammit" phase of development.

arajasek (approved):
Looked at this with fresh-ish eyes, and it seems good.

cgo/errors.go
```diff
@@ -0,0 +1,8 @@
+package cgo
+
+const (
+	ErrInvalidHandle = -1 - iota
```
dignifiedquire, line 4: does this  mean the next one is `0` or `-2`?

dignifiedquire: I would suggest just defining them as their cgo versions, this way these are guranteed to match

cgo/registry.go
dignifiedquire: I meant this one that you wrote: https://github.com/filecoin-project/cgo-blockstore/blob/master/blockstore.go#L17

fvm.go
```diff
@@ -0,0 +1,206 @@
...
+	baseFeeHi, baseFeeLo, err := splitBigInt(opts.BaseFee)
+	if err != nil {
+		return nil, xerrors.Errorf("invalid basefee: %w", err)
+	}
+	baseCircSupplyHi, baseCircSupplyLo, err := splitBigInt(opts.BaseCircSupply)
+	if err != nil {
+		return nil, xerrors.Errorf("invalid circ supply: %w", err)
+	}
+
+	exHandle := cgo.Register(context.TODO(), opts.Externs)
```
dignifiedquire, line 58: pass the context to `CreateFVM`?

bls.go
```diff
@@ -1,8 +1,10 @@
-//+build cgo
+//go:build cgo
```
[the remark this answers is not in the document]
Stebalien: The go build flags format changed, so go now uses _both_ syntaxes (automatically).

fvm.go
Stebalien: Let's reconsider this later. `NewVM` in lotus _does_ take a context, but it's a context for the construction of the VM.

cgo/errors.go
Stebalien: This is -1 as iota starts at 0.

rust/src/fvm/types.rs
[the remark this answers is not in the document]
Stebalien: This is dropped along with the rest of the response. The generated glue is:

```rust
impl Drop for fil_FvmMachineExecuteResponse {
    fn drop(&mut self) {
        unsafe {
            free_c_str(self.error_msg as *mut libc::c_char);
            drop(Vec::from_raw_parts(
                self.return_ptr as *mut u8,
                self.return_len,
                self.return_len,
            ));
        };
    }
}
```

cgo/registry.go
Stebalien: Oh, hell no. Not unless this turns out to be the bottleneck.

rust/src/fvm/types.rs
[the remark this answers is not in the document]
Stebalien: I'm not getting any segfaults, but likely won't wait for the test to limp to a halt.

bls.go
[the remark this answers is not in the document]
dignifiedquire: thanks

cryptonemo (approved):

PR #229, opened 2022-01-28
WIP demo nicer FFI types

Stebalien:
This is a demonstration of nicer FFI types.

1. No more drop macro.
2. No more raw pointers in return types.
3. No more C strings.

Future improvements:

1. Add a Default implementation to FCPResponseStatus so we can derive defaults for all responses.
2. Add a generic response type.

### DIFF
--- a/rust/src/util/types.rs
+++ b/rust/src/util/types.rs
@@ -1,25 +1,115 @@
-use std::ptr;
+use std::{ops::Deref, ptr};
 
 use drop_struct_macro_derive::DropStructMacro;
 // `CodeAndMessage` is the trait implemented by `code_and_message_impl
 use ffi_toolkit::{code_and_message_impl, free_c_str, CodeAndMessage, FCPResponseStatus};
 
 #[repr(C)]
-#[derive(DropStructMacro)]
+pub struct fil_Array<T: Sized> {
+    ptr: *mut T,
+    len: usize,
+}
+
+impl<T> Default for fil_Array<T> {
+    fn default() -> Self {
+        Self {
+            ptr: ptr::null_mut(),
+            len: 0,
+        }
+    }
+}
+
+pub type fil_Bytes = fil_Array<u8>;
+
+impl<T> From<Vec<T>> for fil_Array<T> {
+    fn from(buf: Vec<T>) -> Self {
+        if buf.is_empty() {
+            return Default::default();
+        }
+        buf.into_boxed_slice().into()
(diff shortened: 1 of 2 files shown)


rust/src/util/types.rs
```diff
@@ -1,25 +1,115 @@
...
+
+#[repr(C)]
 pub struct fil_GpuDeviceResponse {
     pub status_code: FCPResponseStatus,
-    pub error_msg: *const libc::c_char,
-    pub devices_len: libc::size_t,
-    pub devices_ptr: *const *const libc::c_char,
+    pub error_msg: fil_Bytes,
+    pub devices: fil_Array<fil_Bytes>,
 }
```
Stebalien, line 105: This should probably be:

```rust
#[repr(C)]
struct fil_Result<T: Sized> {
    pub status_code: FCPResponseStatus,
    pub error_msg: fil_Bytes,
    pub value: T,
}

impl<T> From<Result<T, E>> for fil_Result<T>
where
    T: Sized + Default,
    E: Display,
{
    fn from(r: Result<T, E>) -> Self {
        match r {
            Ok(value) => Self {
                status_code: FCPResponseStatus::FCPNoError,
                error_msg: Default::default(),
                value: value,
            },
            Err(e) => Self {
                status_code: FCPResponseStatus::FCPUnclassifiedError,
                error_msg: e.to_string().into(),
                value: Default::default(),
            },
        }
    }
}
```

rust/src/util/types.rs
```diff
@@ -1,25 +1,115 @@
...
+}
+
+impl<T> From<Box<[T]>> for fil_Array<T> {
+    fn from(buf: Box<[T]>) -> Self {
+        if buf.is_empty() {
+            return Default::default();
+        }
+        let len = buf.len();
+        let ptr = buf.as_mut_ptr();
+        Box::leak(buf);
```
vmx, line 40: Are those to lines the same as `let ptr = Box::into_raw(buf)`?

Stebalien: Not quite. `Box::into_raw(buf)` will return `*const [T]` instead of `*const T`.

vmx: Thanks, I knew I must have missed something.

raulk:
IUC, this has been superseded by the FFI safety refactor that landed here: https://github.com/filecoin-project/filecoin-ffi/pull/247. Reopen if not.

PR #230, opened 2022-01-31
FVM Integration

Stebalien:


Stebalien:
We currently don't have a nice way to tell the FVM to use a fixed circulating supply. Unfortunately, this makes it impossible to run conformance tests from lotus.

Stebalien:
Merged into fvm

PR #235, opened 2022-02-16
fvm: return gasused when verifying consensus faults

arajasek:
Needs https://github.com/filecoin-project/ref-fvm/pull/330

### DIFF
--- a/cgo/extern.go
+++ b/cgo/extern.go
@@ -69,6 +69,7 @@ func cgo_extern_verify_consensus_fault(
 	miner_id *C.uint64_t,
 	epoch *C.int64_t,
 	fault *C.int64_t,
+	gas_used *C.int64_t,
 ) C.int32_t {
 	externs := Lookup(uint64(handle))
 	if externs == nil {
@@ -79,19 +80,15 @@ func cgo_extern_verify_consensus_fault(
 	h2Go := C.GoBytes(unsafe.Pointer(h2), h2_len)
 	extraGo := C.GoBytes(unsafe.Pointer(extra), extra_len)
 
-	res, err := externs.VerifyConsensusFault(context.TODO(), h1Go, h2Go, extraGo)
+	res, gas := externs.VerifyConsensusFault(context.TODO(), h1Go, h2Go, extraGo)
 
-	switch err {
-	case nil:
-		id, err := address.IDFromAddress(res.Target)
-		if err != nil {
-			return ErrIO
-		}
-		*epoch = C.int64_t(res.Epoch)
-		*fault = C.int64_t(res.Type)
-		*miner_id = C.uint64_t(id)
-		return 0
-	default:
+	id, err := address.IDFromAddress(res.Target)
+	if err != nil {
 		return ErrIO
 	}
+	*epoch = C.int64_t(res.Epoch)
+	*fault = C.int64_t(res.Type)
+	*gas_used = C.int64_t(gas)
+	*miner_id = C.uint64_t(id)
+	return 0
 }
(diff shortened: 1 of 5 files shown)


arajasek:
@Stebalien I assume `asr/fvm` is functionally the fvm dev branch by this point?

rust/Cargo.toml
```diff
@@ -74,8 +74,8 @@ bindgen=[]
 # Patched in
 [patch.crates-io]
 cid = { git = "https://github.com/multiformats/rust-cid", branch = "steb/cbor-hack" }
-fvm = { git = "https://github.com/filecoin-project/ref-fvm", branch = "steb/extern" }
-fvm_shared = { git = "https://github.com/filecoin-project/ref-fvm", branch = "steb/extern" }
+fvm = { path = '../../../../fvm/fvm' }
```
arajasek, line 77: Drop after https://github.com/filecoin-project/ref-fvm/pull/330 lands, but not sure what to patch to? master? no patch at all?

Stebalien: I think we can drop the patch entirely and just depend on the git version.

cgo/extern.go
```diff
@@ -69,6 +69,7 @@ func cgo_extern_verify_consensus_fault(
 	miner_id *C.uint64_t,
 	epoch *C.int64_t,
 	fault *C.int64_t,
+	gas_used *C.int64_t,
```
Stebalien, line 72: Given that we're already using an "out" pointer here, what if we passed in a pointer to some `gas_remaining` variable? That would let us stop early without changing the parameters. I.e.:

1. Read the current gas remaining from the `gas_remaining` parameter.
2. Update it locally, stopping if/when we run out.
3. Write back the final value.

cgo/interface.go
```diff
@@ -30,7 +31,7 @@ const (
 type Externs interface {
 	GetChainRandomness(ctx context.Context, personalization crypto.DomainSeparationTag, epoch abi.ChainEpoch, entropy []byte) ([]byte, error)
 	GetBeaconRandomness(ctx context.Context, personalization crypto.DomainSeparationTag, epoch abi.ChainEpoch, entropy []byte) ([]byte, error)
-	VerifyConsensusFault(ctx context.Context, h1, h2, extra []byte) (*ConsensusFault, error)
+	VerifyConsensusFault(ctx context.Context, h1, h2, extra []byte) *ConsensusFaultWithGas
```
Stebalien, line 34: Maybe just return multiple values?

Stebalien: (`ConsensusFaultWithGas` is a bit of a strange type)

rust/src/fvm/externs.rs
```diff
@@ -99,17 +99,19 @@ impl Consensus for CgoExterns {
                 &mut miner_id,
                 &mut epoch,
                 &mut fault_type,
+                &mut gas_used,
             ) {
-                0 => Ok(Some(ConsensusFault {
-                    target: Address::new_id(miner_id),
+                0 => Ok(ConsensusFaultWithGas {
```
Stebalien, line 104: Personally, I'd return a tuple instead of a new type.

Also, this was clearly broken before as we had no case that returned `None`, but I don't think that having a "no fault" variant is the right fix. See https://github.com/filecoin-project/ref-fvm/pull/330#discussion_r807523704.

Stebalien (changes requested):

cgo/extern.go
Stebalien: Although I guess this might complicate some other parts of the API a bit?

Stebalien (approved):

PR #236, opened 2022-02-16
fix: specify cargo build directory explicitly

Aloxaf:
Some users may change the default value of [build.target-dir](https://doc.rust-lang.org/cargo/reference/config.html#buildtarget-dir), which will cause this project failed to build.


Stebalien (approved):
I'm generally not a fan of overriding the user's choice... but I can't see a better solution.

PR #242, opened 2022-03-11
fix: allow fake code CIDs

Stebalien:


arajasek (approved):
This is pretty clean honestly

PR #245, opened 2022-03-15
Fix fil_fvm_machine_flush (const)

turuslan:
Can't call `fil_destroy_fvm_machine_flush_response` from C because of type mismatch.  
Removing `const` to fix.

```diff
// generated C header
-const struct fil_FvmMachineFlushResponse *fil_fvm_machine_flush(void *executor);
+struct fil_FvmMachineFlushResponse *fil_fvm_machine_flush(void *executor);
void fil_destroy_fvm_machine_flush_response(struct fil_FvmMachineFlushResponse *ptr);
```

### DIFF
--- a/rust/src/fvm/machine.rs
+++ b/rust/src/fvm/machine.rs
@@ -212,7 +212,7 @@ pub unsafe extern "C" fn fil_fvm_machine_execute_message(
 #[no_mangle]
 pub unsafe extern "C" fn fil_fvm_machine_flush(
     executor: *mut libc::c_void,
-) -> *const fil_FvmMachineFlushResponse {
+) -> *mut fil_FvmMachineFlushResponse {
     catch_panic_response(|| {
         init_log();
 


turuslan:
`fil_destroy_fvm_machine_flush_response` expects `*mut` pointer, but `fil_fvm_machine_flush` returns `*const` pointer.
`*const` argument can't be passed to `*mut` parameter.

`const` is not needed here, all other ffi functions return `*mut`.
```diff
pub unsafe extern "C" fn fil_fvm_machine_flush(
    executor: *mut libc::c_void,
-) -> *const fil_FvmMachineFlushResponse {
+) -> *mut fil_FvmMachineFlushResponse {
...
pub unsafe extern "C" fn fil_destroy_fvm_machine_flush_response(
    ptr: *mut fil_FvmMachineFlushResponse,
) {
```

Stebalien:
Oh, hm. I see. I wonder why go wasn't complaining...

Issue #231, opened 2022-01-31
Enums in FFI types are unsafe


Stebalien:
It looks like we're using enums in FFI types in several places, but this is inherently unsafe as we can end up creating enums with arbitrary values and the rust type system will assume that this can't happen.

This can cause problems like:

1. Code checking said enum value may be optimized away.
2. If the enum has no zero variant but a zero is passed, rust may interpret `Option<Enum>` as `None`.
3. Really, whatever the rust compiler wants to do.

dignifiedquire:
Asfaict all enums that are used are using `repr(C)` which should make them FFI safe, so I don't think these concerns apply.

Stebalien:
It's unsafe because we call into the FFI from go without checking whether the enum variant is valid. That means we could end up constructing an "impossible" enum variant (e.g., 9999, when the only valid enum variants are 0-5).

Given that the rust compiler assumes that enum variants are statically "valid" without checking, this could, e.g., cause a match statement (if implemented as a jump table) to jump to some random offset in memory.

Stebalien:
Fixed.

PR #247, opened 2022-03-18
fix: improve ffi safety

dignifiedquire:
- *mut ptr to indicate ownership
- Boxed slices instead of vecs for safe transfers
- expanded checks for null ptrs
- refactor to use no more CStrings
- improve api and safety
- `C -> go bindings`: remove c-for-go in favor of manual bindings
- `rust -> C bindings`: use [safer_ffi](https://docs.rs/safer-ffi/latest/safer_ffi/) instead of bindgen
- check for enum values passed in the ffi interface

Closes https://github.com/filecoin-project/rust-fil-ffi-toolkit/issues/9, #231 

TODOs

- [x] update low level go bindings
- [x] update high level go bindings
- [x] green CI
- [x] Apply CR
- [x] Cleanup pass
- [x] Rebase on master with fvm integration
- [x] Update fvm bindings (rust)
- [x] Update fvm bindings(go)

~~Depends on https://github.com/filecoin-project/rust-fil-ffi-toolkit/pull/11~~


rust/src/util/types.rs
```diff
@@ -1,55 +1,279 @@
...
 
 #[no_mangle]
 pub unsafe extern "C" fn fil_destroy_init_log_fd_response(ptr: *mut fil_InitLogFdResponse) {
     let _ = Box::from_raw(ptr);
 }
+
+pub fn vec_into_raw<T>(v: Vec<T>) -> (*mut T, usize) {
+    let bytes = v.into_boxed_slice();
+    let len = bytes.len();
+    (Box::into_raw(bytes).cast(), len)
```
Stebalien, line 238: So... if this vec is empty, the pointer will be a pointer to `1`, not null. If we pass that to _go_, and try to construct a slice from that, go will segfault (speaking from experience).

rust/src/util/types.rs
```diff
@@ -1,55 +1,279 @@
...
+/// `ptr` must be allocated using `Box::into_raw` or be null.
+pub unsafe fn clone_box_parts<T: Clone>(ptr: *const T, len: usize) -> *mut T {
+    if ptr.is_null() {
+        return ptr::null_mut();
+    }
+
+    // restore, but without triggering a drop on the original
+    // safe to cast as *mut as we don't mutate
+    let bytes: ManuallyDrop<Box<[T]>> = ManuallyDrop::new(Box::from_raw(
+        std::slice::from_raw_parts_mut(ptr as *mut _, len),
```
Stebalien, line 257: This is likely unsound as it'll create a mutable reference (and this likely isn't the only reference).

rust/src/proofs/api.rs
```diff
@@ -192,7 +192,7 @@ pub unsafe extern "C" fn fil_fauxrep2(
             }
             Err(err) => {
                 response.status_code = FCPResponseStatus::FCPUnclassifiedError;
-                response.error_msg = rust_str_to_c_str(format!("{:?}", err));
+                response.error_msg = rust_str_to_c_str(format!("{:?}", err)).unwrap();
```
Stebalien, line 195: Ideally we'd either:

1. Store these as "byte" strings (ptr & length) instead of C strings, given that we don't really care about C.
2. Escape null bytes. I.e., have a `rust_err_to_c` function that converts errors to strings, escaping null bytes, special characters, etc.

Stebalien: Or is the goal here to use the `fil_Result` type?

dignifiedquire: this will go away soon

rust/src/util/types.rs
```diff
@@ -1,55 +1,279 @@
...
+    fn drop(&mut self) {
+        unsafe { drop_box_from_parts(self.ptr) }
+    }
+}
+
+impl<T> Deref for fil_Array<T> {
+    type Target = [T];
+    fn deref(&self) -> &[T] {
+        unsafe {
+            if self.ptr.is_null() {
```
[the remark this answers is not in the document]
dignifiedquire: fixed

rust/src/util/types.rs
```diff
@@ -1,55 +1,279 @@
...
+    }
+
+    // restore, but without triggering a drop on the original
+    // safe to cast as *mut as we don't mutate
+    let bytes: ManuallyDrop<Box<[T]>> = ManuallyDrop::new(Box::from_raw(
+        std::slice::from_raw_parts_mut(ptr as *mut _, len),
+    ));
+
+    // duplicate the bytes
+    bytes.to_vec()
```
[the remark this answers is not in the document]
dignifiedquire: fixed

rust/src/util/types.rs
```diff
@@ -1,55 +1,279 @@
...
+
+    // restore, but without triggering a drop on the original
+    // safe to cast as *mut as we don't mutate
+    let bytes: ManuallyDrop<Box<[T]>> = ManuallyDrop::new(Box::from_raw(
+        std::slice::from_raw_parts_mut(ptr as *mut _, len),
+    ));
+
+    // duplicate the bytes
+    let bytes2 = ManuallyDrop::into_inner(bytes.clone());
+    Box::into_raw(bytes2).cast()
```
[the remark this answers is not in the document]
dignifiedquire: fixed

rust/src/util/types.rs
dignifiedquire: fixed

rust/src/util/types.rs
dignifiedquire: removed

rust/src/proofs/api.rs
```diff
@@ -45,48 +42,29 @@ pub unsafe extern "C" fn fil_write_with_alignment(
...
-                response.left_alignment_unpadded = (written - n).into();
-                response.status_code = FCPResponseStatus::FCPNoError;
-                response.total_write_unpadded = written.into();
-            }
-            Err(err) => {
-                response.status_code = FCPResponseStatus::FCPUnclassifiedError;
-                response.error_msg = rust_str_to_c_str(format!("{:?}", err));
-            }
-        }
+        )?;
```
Stebalien, line 61: :smile: This is so much nicer!

rust/src/proofs/api.rs
```diff
@@ -100,105 +78,51 @@ pub unsafe extern "C" fn fil_write_without_alignment(
...
     })
 }
 
 #[no_mangle]
 pub unsafe extern "C" fn fil_fauxrep(
     registered_proof: fil_RegisteredSealProof,
-    cache_dir_path: *const libc::c_char,
-    sealed_sector_path: *const libc::c_char,
+    cache_dir_path: fil_Bytes,
+    sealed_sector_path: fil_Bytes,
```
Stebalien, line 100: `fil_Bytes` will free the underlying memory. Do we need to some form of "borrowed" form of this type that doesn't drop? Not sure if we own these paths or not.

dignifiedquire: for now I am making them all owned, but probably should add a ref version and optimize at least the paths

dignifiedquire: looks like I can just use `&fil_Bytes`

rust/src/proofs/api.rs
```diff
@@ -655,183 +379,101 @@ pub unsafe extern "C" fn fil_verify_seal(
...
-                            challenges_stride_mismatch = true;
-                        }
-
-                        challenges
-                    })
-                    .collect();
+        let sector_ids: Vec<u64> = output
+            .clone()
+            .into_iter()
+            .map(|(id, _challenges)| u64::from(id))
```
Stebalien, line 429: nit:

```suggestion
            .map(|(id, _)| u64::from(id))
```

rust/src/util/types.rs
```diff
@@ -1,55 +1,325 @@
...
+            let _ = unsafe { Box::from_raw(ptr.as_ptr()) };
+        }
+    }
+}
+
+impl<T> Deref for fil_Array<T> {
+    type Target = [T];
+    fn deref(&self) -> &[T] {
+        match self.ptr {
+            None => unsafe { std::slice::from_raw_parts(ptr::NonNull::dangling().as_ptr(), 0) },
```
Stebalien, line 130: nit: I _think_ you can write this as `&[]`, right?

rust/src/util/types.rs
```diff
@@ -1,55 +1,325 @@
...
+    }
+}
+
+impl fil_Bytes {
+    pub fn as_str(&self) -> Result<&str, Utf8Error> {
+        std::str::from_utf8(self)
+    }
+
+    pub fn as_path(&self) -> Result<PathBuf, Utf8Error> {
+        self.as_str().map(PathBuf::from)
```
Stebalien, line 184: Are we even going to pretend to support windows? If not... we can use https://doc.rust-lang.org/std/os/unix/ffi/trait.OsStrExt.html.

Stebalien: If yes, let me introduce you to utf16.

Stebalien (commented):
A few more comments, but otherwise this looks good to go.

rust/src/util/types.rs
dignifiedquire: you wrote this :P, haven't tried it yet

rust/src/util/types.rs
dignifiedquire: fixed

rust/src/proofs/api.rs
dignifiedquire: fixed

rust/src/proofs/api.rs
```diff
@@ -18,80 +15,55 @@ use filecoin_proofs_api::{
...
 #[no_mangle]
 #[cfg(not(target_os = "windows"))]
 pub unsafe extern "C" fn fil_write_with_alignment(
     registered_proof: fil_RegisteredSealProof,
     src_fd: libc::c_int,
     src_size: u64,
     dst_fd: libc::c_int,
-    existing_piece_sizes_ptr: *const u64,
-    existing_piece_sizes_len: libc::size_t,
+    existing_piece_sizes: fil_Array<u64>,
```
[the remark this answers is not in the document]
dignifiedquire: fixed

rust/src/util/types.rs
dignifiedquire: fixed

rust/src/util/types.rs
```diff
@@ -1,55 +1,325 @@
...
+        fil_Result::from(callback().map_err(|err| err.to_string()))
+    })
+}
+
+pub fn catch_panic_response_raw<F, T>(name: &str, callback: F) -> *mut fil_Result<T>
+where
+    T: Sized + Default,
+    F: FnOnce() -> fil_Result<T>,
+{
+    // Using AssertUnwindSafe is code smell. Though catching our panics here is really
```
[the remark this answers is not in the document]
dignifiedquire: removed it, looks like we are okay for now

rust/src/proofs/api.rs
```diff
@@ -100,377 +72,193 @@ pub unsafe extern "C" fn fil_write_without_alignment(
...
     registered_aggregation: fil_RegisteredAggregationProof,
-    comm_rs_ptr: *const fil_32ByteArray,
-    comm_rs_len: libc::size_t,
-    seeds_ptr: *const fil_32ByteArray,
-    seeds_len: libc::size_t,
-    seal_commit_responses_ptr: *const fil_SealCommitPhase2Response,
-    seal_commit_responses_len: libc::size_t,
+    comm_rs: fil_Array<fil_32ByteArray>,
+    seeds: fil_Array<fil_32ByteArray>,
+    seal_commit_responses: fil_Array<fil_SealCommitPhase2>,
```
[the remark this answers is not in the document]
dignifiedquire: fixed

rust/src/proofs/api.rs
```diff
@@ -100,377 +72,193 @@ pub unsafe extern "C" fn fil_write_without_alignment(
...
 
 /// TODO: document
-///
 #[no_mangle]
 pub unsafe extern "C" fn fil_seal_pre_commit_phase2(
-    seal_pre_commit_phase1_output_ptr: *const u8,
-    seal_pre_commit_phase1_output_len: libc::size_t,
-    cache_dir_path: *const libc::c_char,
-    sealed_sector_path: *const libc::c_char,
+    seal_pre_commit_phase1_output: fil_Bytes,
```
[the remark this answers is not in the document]
dignifiedquire: fixed

rust/src/bls/api.rs
```diff
@@ -88,10 +93,14 @@ pub unsafe extern "C" fn fil_hash(
 ///
 /// Returns `NULL` on error. Result must be freed using `destroy_aggregate_response`.
 #[no_mangle]
-pub unsafe extern "C" fn fil_aggregate(
+pub unsafe extern "C" fn aggregate(
     flattened_signatures_ptr: *const u8,
     flattened_signatures_len: libc::size_t,
-) -> *mut types::fil_AggregateResponse {
+) -> *mut types::AggregateResponse {
```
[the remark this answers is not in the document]
dignifiedquire: done

magik6k (commented):
Not seeing anything very wrong here, but I feel like this probably deserves at least one more look from someone who in more into rust.

cgo/types.go
```diff
@@ -0,0 +1,530 @@
...
+	"unsafe"
+)
+
+type FCPResponseStatus int64
+
+type SliceBoxedUint8 = C.struct_slice_boxed_uint8
+
+func (ptr SliceBoxedUint8) Slice() []byte {
+	return unsafe.Slice((*byte)(ptr.ptr), int(ptr.len))
+}
```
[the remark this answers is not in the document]
Stebalien: I just want to make it _super_ clear that the underlying slice may be freed.

cgo/proofs.go
```diff
@@ -0,0 +1,333 @@
...
+/*
+#cgo LDFLAGS: -L${SRCDIR}/..
+#cgo pkg-config: ${SRCDIR}/../filcrypto.pc
+#include "../filcrypto.h"
+#include <stdlib.h>
+*/
+import "C"
+
+func VerifySeal(registeredProof RegisteredSealProof, commR *ByteArray32, commD *ByteArray32, proverId *ByteArray32, ticket *ByteArray32, seed *ByteArray32, sectorId uint64, proof SliceRefUint8) (bool, error) {
+	resp := C.verify_seal(registeredProof, commR, commD, proverId, ticket, seed, C.uint64_t(sectorId), proof)
```
Stebalien, line 12: We can do it in a followup, but we should be checking that, e.g., `RegisteredSealProof` is a valid value for the rust enum.

cgo/types.go
```diff
@@ -0,0 +1,530 @@
...
+		return []byte{}
+	}
+
+	res := make([]byte, int(ptr.len))
+	copy(res, ptr.Slice())
+	return res
+}
+
+func (ptr SliceBoxedUint8) Destroy() {
+	C.destroy_boxed_slice(ptr)
```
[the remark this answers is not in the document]
Stebalien: Alternatively... what if we allocated `*SliceBoxedUint8` and used a finalizer? Too expensive?

We can explore it later, I'm just trying to find ways to, well, have fewer potential bugs.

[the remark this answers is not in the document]
Stebalien: Of course, if we used a finalizer, we'd have to be _very_ careful about things like `ptr.Slice`.

cgo/types.go
```diff
@@ -0,0 +1,530 @@
...
+
+func (ptr *BLSDigest) Slice() []byte {
+	return ptr.inner.Slice()
+}
+
+type HashResponse = C.HashResponse_t
+
+func (ptr *HashResponse) Digest() []byte {
+	return ptr.digest.Slice()
+}
```
Stebalien, line 48: For stuff like this, I'd consider either:

1. Copying.
2. Returning the "desired" type (in this case, `Digest`).

Stebalien (commented):
Some ideas/comments. Honestly, the unsafety of things like `ptr.Slice()` is probably fine given the context.

cgo/proofs.go
dignifiedquire: our new ffi bindings generate checks for this now https://getditto.github.io/safer_ffi/ffi-export/sanity-checks.html

cgo/types.go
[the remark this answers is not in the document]
dignifiedquire: these are sometimes managed on the rust side sometimes on the go side, so this needs to be manual

cgo/types.go
```diff
@@ -0,0 +1,530 @@
...
+#include <stdlib.h>
+*/
+import "C"
+import (
+	"unsafe"
+)
+
+type FCPResponseStatus int64
+
+type SliceBoxedUint8 = C.struct_slice_boxed_uint8
```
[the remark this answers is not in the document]
dignifiedquire: we could, but this way it is easier to match to the other lang side and not get confused

cgo/types.go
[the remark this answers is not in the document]
Kubuxu: Also, you should not rely on the finalizer on the go side to ever run. If Go doesn't see memory pressure within its own GC tracked area, it won't run the finalizer.

cgo/proofs.go
Stebalien: Oh that's wonderful.

cgo/types.go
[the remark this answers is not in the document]
Stebalien: Hm. I wonder we want different types then. But I don't feel strongly about this as long as it's clear when/where Destroy needs to be called.

cgo/types.go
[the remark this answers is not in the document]
Stebalien: I guess. It's just... ew.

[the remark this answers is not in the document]
dignifiedquire: welcome to this repo

rust/src/bls/api.rs
```diff
@@ -181,39 +167,35 @@ pub unsafe extern "C" fn fil_verify(
...
 /// # Arguments
 ///
-/// * `signature_ptr`             - pointer to a signature byte array (SIGNATURE_BYTES long)
-/// * `messages_ptr`              - pointer to an array containing the pointers to the messages
+/// * `signature`             - pointer to a signature byte array (SIGNATURE_BYTES long)
+/// * `messages`              - pointer to an array containing the pointers to the messages
 /// * `messages_sizes_ptr`        - pointer to an array containing the lengths of the messages
 /// * `messages_len`              - length of the two messages arrays
 /// * `flattened_public_keys_ptr` - pointer to a byte array containing public keys
 /// * `flattened_public_keys_len` - length of the array
```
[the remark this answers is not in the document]
vmx:
```suggestion
/// * `signature`             - byte array containing the signature
/// * `messages`              - array containing the pointers to the messages
/// * `messages_sizes`        - array containing the lengths of the messages
/// * `flattened_public_keys` - byte array containing public keys
```

rust/src/bls/api.rs
```diff
@@ -181,39 +166,35 @@ pub unsafe extern "C" fn fil_verify(
...
 ) -> libc::c_int {
     // prep request
-    let raw_signature = from_raw_parts(signature_ptr, SIGNATURE_BYTES);
-    let signature = try_ffi!(Signature::from_bytes(raw_signature), 0);
+    let signature = try_ffi!(Signature::from_bytes(&signature), 0);
 
-    let flattened = from_raw_parts(flattened_messages_ptr, flattened_messages_len);
-    let chunk_sizes = from_raw_parts(message_sizes_ptr, message_sizes_len);
+    let flattened = flattened_messages;
+    let chunk_sizes = message_sizes;
```
vmx, line 186: I'd remove those, unless you think naming it `chunk_sizes` makes things clearer (I don't think so).

rust/src/bls/api.rs
```diff
@@ -181,39 +166,35 @@ pub unsafe extern "C" fn fil_verify(
...
-    let mut messages: Vec<&[u8]> = Vec::with_capacity(message_sizes_len);
+    let mut messages: Vec<&[u8]> = Vec::with_capacity(chunk_sizes.len());
     let mut offset = 0;
     for chunk_size in chunk_sizes.iter() {
         messages.push(&flattened[offset..offset + *chunk_size]);
         offset += *chunk_size
     }
 
-    let raw_public_keys = from_raw_parts(flattened_public_keys_ptr, flattened_public_keys_len);
+    let raw_public_keys = flattened_public_keys;
```
vmx, line 197: I'd also remove this assignment.

rust/src/bls/api.rs
```diff
@@ -55,105 +57,88 @@ macro_rules! try_ffi {
...
         .write_bytes(&mut raw_signature.as_mut())
         .expect("preallocated");
 
-    let response = types::fil_AggregateResponse {
-        signature: fil_BLSSignature {
+    let response = types::AggregateResponse {
+        signature: BLSSignature {
             inner: raw_signature,
         },
     };
```
vmx, line 108: Just an idea, I don't know if that's Rustic or not (or a good idea). Could this perhaps be something like:

```rust
let signature = BLSSignature::new();
aggregated
    .write_bytes(&mut signature.as_mut())
    .expect("preallocated");
```

The same would then apply to the other BLS things.

rust/src/bls/api.rs
```diff
@@ -256,61 +237,53 @@ pub unsafe extern "C" fn fil_private_key_generate() -> *mut types::fil_PrivateKe
...
 
 /// Sign a message with a private key and return the signature
 ///
 /// # Arguments
 ///
-/// * `raw_private_key_ptr` - pointer to a private key byte array
-/// * `message_ptr` - pointer to a message byte array
-/// * `message_len` - length of the byte array
+/// * `raw_private_key` - pointer to a private key byte array
+/// * `message` - pointer to a message byte array
```
vmx, line 265: Those are not pointers anymore (at least you didn't use the word "pointer" in other cases).

vmx (commented):
Only small things and some of them that could/should be done on subsequent PRs.

cgo/types.go
```diff
@@ -0,0 +1,530 @@
...
+
+func (ptr SliceBoxedUint8) Slice() []byte {
+	return unsafe.Slice((*byte)(ptr.ptr), int(ptr.len))
+}
+
+func (ptr SliceBoxedUint8) Copy() []byte {
+	if ptr.len == 0 {
+		return []byte{}
```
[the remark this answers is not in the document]
dignifiedquire: fixed

rust/src/bls/api.rs
dignifiedquire: fixed

rust/src/bls/api.rs
```diff
@@ -55,105 +57,88 @@ macro_rules! try_ffi {
...
 
-    Box::into_raw(Box::new(response))
+    Some(repr_c::Box::new(response))
 }
 
 /// Verify that a signature is the aggregated signature of hashes - pubkeys
 ///
 /// # Arguments
 ///
 /// * `signature_ptr`             - pointer to a signature byte array (SIGNATURE_BYTES long)
```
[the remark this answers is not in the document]
dignifiedquire: fixed

rust/src/bls/api.rs
dignifiedquire: fixed

rust/src/bls/api.rs
dignifiedquire: fixed

cgo/types.go
dignifiedquire: cleaned up

cgo/types.go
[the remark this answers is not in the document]
dignifiedquire: added an additonal check for the inner ptr to not be null

rust/src/bls/api.rs
vmx: Wow that's really nice and much shorter now.

vmx (approved):
I'm still not a fan of glob imports, but other than that it looks good. I've only reviewed the Rust part of it btw.

rust/src/fvm/blockstore/cgo.rs
```diff
@@ -34,13 +34,16 @@ impl Blockstore for CgoBlockstore {
             match cgo_blockstore_has(self.handle, k_bytes.as_ptr(), k_bytes.len() as i32) {
                 // We shouldn't get an "error not found" here, but there's no reason to be strict
                 // about it.
-                0 | ERR_NOT_FOUND => Ok(false),
+                0 => Ok(false),
+                x if x == FvmError::NotFound as i32 => Ok(false),
```
vmx, line 38: Nitpick (this doesn't prevent merging and could be done later on). In this match the thing it matches on is quite often bound to a variable. I think binding it directly would make the code easier to read. Especially since the match is a function (in the cases below with quite a few parameters). It would also reduce the scope of `unsafe`. I propose something like:

```rust
let has = unsafe { cgo_blockstore_has(self.handle, k_bytes.as_ptr(), k_bytes.len() as i32) } as i32;
match has {
    // We shouldn't get an "error not found" here, but there's no reason to be strict
    // about it.
    0 => Ok(false),
    FvmError::NotFound => Ok(false),
    1 => Ok(true),
    // Panic on unknown values. There's a bug in the program.
    2.. => panic!("invalid return value from has: {}", has),
    // Panic if the store isn't registered. This means something _very_ unsafe is going
    // on and there is a bug in the program.
    FvmError::InvalidHandle => {
        panic!("blockstore {} not registered", self.handle)
    }
    // Otherwise, return "other". We should add error codes in the future.
    _ => Err(anyhow!("cgo blockstore 'has' failed with error code {}", has)),
}
```

Though it's a matter of style and not correctness, so feel free to ignore.

rust/src/fvm/machine.rs
```diff
@@ -31,265 +34,173 @@ lazy_static! {
...
-
-        let network_version = match NetworkVersion::try_from(network_version as u32) {
-            Ok(x) => x,
-            Err(_) => {
-                response.status_code = FCPResponseStatus::FCPUnclassifiedError;
-                response.error_msg =
-                    rust_str_to_c_str(format!("unsupported network version: {}", network_version));
-                return raw_ptr(response);
+) -> repr_c::Box<Result<repr_c::Box<FvmMachine>>> {
+    unsafe {
```
vmx, line 51: Is the `unsafe` block needed here?

vmx (approved):
Only two minor, non blocking things.

I've only reviewed the Rust side of things. Great work! This is really a massive improvement and also a huge amount of tiresome work. Wow.

rust/src/fvm/machine.rs
dignifiedquire: yes, because it uses an `unsafe` version of the panic catcher

rust/src/fvm/blockstore/cgo.rs
dignifiedquire: I agree this code could be cleaned up, I was going for minimal changes in here this time around 😅  given the amount of code this PR already changes.

Kubuxu:
I will restate the question regarding changes to function signatures. filecoin-ffi is used by multiple downstream projects, and changes to function signatures will have painful side-effects.

cgo/helpers.go
```diff
@@ -0,0 +1,287 @@
...
+	"unsafe"
+)
+
+func AsSliceRefUint8(goBytes []byte) SliceRefUint8 {
+	len := len(goBytes)
+
+	if len == 0 {
+		// can't take element 0 of an empty slice
+		return SliceRefUint8{
+			ptr: (*C.uint8_t)(unsafe.Pointer(&goBytes)),
```
Kubuxu, line 21: I would feel better if we had a global `uint8` variable and were taking pointer to it instead of the slice.
Reinterpreting the slice pointer to `C.uint8_t` feels wrong.

Kubuxu: But it would make the code more verbose (all different types).
Alternative would be the pointer to be `nil`.

dignifiedquire:
> I will restate the question regarding changes to function signatures. filecoin-ffi is used by multiple downstream projects, and changes to function signatures will have painful side-effects.

Can you clarify which part of the api. The C-API or the go-api?

dignifiedquire:
> Same comment as previously on Slack: as the bindings are written for specific headers, I would would prefer if header file itself was checked in. I can be convinced otherwise but from my perspective, it makes changes more auditable as most Go types not are aliases to C types.

I don't mind either way, happy to do as others think makes the most sense

cgo/helpers.go
dignifiedquire: I don't understand the concern/solution I am afraid.

Kubuxu:
> Can you clarify which part of the api. The C-API or the go-api?

I know of few projects using the Go side of filecoin-ffi. I'm not aware of any using the C/rust side.

dignifiedquire:
> I know of few projects using the Go side of filecoin-ffi. I'm not aware of any using the C/rust side.

Then that should be fine, I need to do a review, but the actual go api should not have changed much if at all.

Kubuxu:
From what I saw there are few changes to bls APIs, so if we can keep them the same without any issues we should.
If there is good reason to change them, I wouldn't be pushing back much.

dignifiedquire:
@Kubuxu undid the public api changes, they would be nice to do for a cleaner api, but not necessary now

cryptonemo (approved):
This is a lot of change, but generally looks much cleaner. Thanks

cgo/helpers.go
raulk: @Kubuxu can you re-explain this if it's still a concern?

raulk:
Seems like there's sufficient buy-in now to merge this refactor. There's a proposed course of action to land it on master:

1. Resolve merge conflicts with master. @dignifiedquire -- are you able to own this?
2. Fully resolve @Kubuxu's comments:
    - Breaking changes in public Go API: with the reversal of the BLS interface change above, do we have any further concerns?
    - Close this unresolved thread: https://github.com/filecoin-project/filecoin-ffi/pull/247#discussion_r849596753
3. Final review from @Stebalien.

cgo/helpers.go
dignifiedquire: explained in person, fix just pushed

raulk:
Unfortunately this requires a merge from master again following the merge of execution traces. @dignifiedquire do you have a few minutes to look at this?

Stebalien (approved):

cgo/types.go
```diff
@@ -617,6 +617,6 @@ func (r FvmMachineExecuteResponse) copy() FvmMachineExecuteResponseGo {
 		MinerTipHi:  uint64(r.miner_tip_hi),
 		MinerTipLo:  uint64(r.miner_tip_lo),
 		ExecTrace:   r.exec_trace.copy(),
-		FailureInfo: string(r.failure_info.copy()),
+		FailureInfo: string(r.failure_info.slice()),
```
dignifiedquire, line 620: are you sure that `string()` copies?

cgo/types.go
```diff
@@ -0,0 +1,622 @@
...
+	}
+}
+
+func (ptr *resultSealPreCommitPhase2) statusCode() FCPResponseStatus {
+	return FCPResponseStatus(ptr.status_code)
+}
+
+func (ptr *resultSealPreCommitPhase2) errorMsg() *SliceBoxedUint8 {
+	return &ptr.error_msg
+}
```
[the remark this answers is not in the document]
dignifiedquire: the convention is, anything in a public interface of cgo is owned, borrows are only allowed in the cgo internals

proofs.go
```diff
@@ -805,402 +659,329 @@ func FauxRep2(proofType abi.RegisteredSealProof, cacheDirPath string, existingPA
...
 
-func toFilPrivateReplicaInfos(src []PrivateSectorInfo, typ string) ([]generated.FilPrivateReplicaInfo, uint, func(), error) {
-	allocs := make([]AllocationManager, len(src))
+func makeCleanerPRI(src []cgo.PrivateReplicaInfo, limit int) func() {
+	return func() {
+		for i := 0; i < limit; i++ {
+			src[i].Destroy()
+		}
+	}
+}
```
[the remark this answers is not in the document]
dignifiedquire: sorry, not sure what you mean

distributed.go
```diff
@@ -106,26 +88,19 @@ func GenerateWinningPoStWithVanilla(
...
-	resp.Deref()
-	resp.ProofsPtr = make([]generated.FilPoStProof, resp.ProofsLen)
-	resp.Deref()
-
-	defer generated.FilDestroyGenerateWinningPostResponse(resp)
+	fproofs, cleanup, err := toVanillaProofs(proofs)
+	defer cleanup()
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
dignifiedquire: agreed, fixed

cgo/types.go
Stebalien: It's supposed to? I think go will only optimize this in _very_ special cases. Basically, where it can inline things.

proofs.go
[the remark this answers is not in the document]
Stebalien:
```go
for i := limit - 1; i >= 0; i--
```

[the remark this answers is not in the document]
Stebalien: But almost certainly not worth it (and harder to read).

rust/src/fvm/types.rs
```diff
@@ -14,10 +14,12 @@ pub enum FvmRegisteredVersion {
 #[derive_ReprC]
 #[ReprC::opaque]
 #[derive(Default)]
-pub struct FvmMachine {
+pub struct InnerFvmMachine {
     pub(crate) machine: Option<Mutex<CgoExecutor>>,
 }
 
+pub type FvmMachine = Option<repr_c::Box<InnerFvmMachine>>;
```
Stebalien, line 21: This is a very confusing type.

cgo/types.go
Kubuxu: I'm 99% sure it doesn't.

dignifiedquire: then this needs to be reverted

cgo/proofs.go
```diff
@@ -169,13 +169,13 @@ func GenerateWinningPoSt(randomness *ByteArray32, replicas SliceRefPrivateReplic
 func GenerateWindowPoSt(randomness *ByteArray32, replicas SliceRefPrivateReplicaInfo, proverId *ByteArray32) ([]PoStProofGo, []uint64, error) {
 	resp := C.generate_window_post(randomness, replicas, proverId)
 	defer resp.destroy()
-	faults := resp.value.faulty_sectors.copy()
-
 	if err := CheckErr(resp); err != nil {
+		faults := resp.value.faulty_sectors.copy()
```
Stebalien, line 173: This is invalid memory otherwise? Maybe we should fix it in rust?

cgo/types.go
Stebalien: I'm 99.99999999999% sure it gets copied.

cgo/proofs.go
dignifiedquire: it wasn’t an issue, just wanted to make the code more defensive on both sides

rust/src/fvm/types.rs
dignifiedquire: not a great type, but necessary, I'll add a comment

proofs.go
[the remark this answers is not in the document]
dignifiedquire: yeah

cgo/types.go
Kubuxu: Ok, then it is the 1%, I thought it was ‘string(string)’ cast.

rust/scripts/build-release.sh
```diff
@@ -64,6 +64,10 @@ main() {
         find . -type f -name "lib$1.a"
     fi
 
+    # generate filcrypto.h
+    RUSTFLAGS="${__rust_flags}" HEADER_DIR="." \
+        cargo test build_headers --features c-headers
```
Stebalien, line 69: @dignifiedquire why build with a test?

Stebalien: Basically, it means we need to rebuild _everything_ (and we should just be using `cargo run` with a bin target).

Stebalien:
Ok. I've now synced the chain with the address sanitizer enabled (both in rust and go) with both the FVM and the lotus VM.

Now I'm running the lotus tests with the address sanitizer enabled.

rust/scripts/build-release.sh
dignifiedquire: Mostly because that is how the `safer-ffi` docs do it, and it didn't quite work when doing it differently due to how that project currently works.

Stebalien:
I'm getting:

```
    github.com/filecoin-project/lotus/storage.(*WindowPoStScheduler).runPoStCycle
        /var/lib/lotus/src/storage/wdpost_run.go:725
  - merge windowPoSt partition proofs:
    github.com/filecoin-project/lotus/extern/sector-storage.(*Manager).generateWindowPoSt
        /var/lib/lotus/src/extern/sector-storage/manager_post.go:207
  - SlicedBoxedUint8 must not be empty
```

Why can this not be empty?

laser:
> C -> go bindings: remove c-for-go in favor of manual bindings

Wow!

PR #249, opened 2022-03-24
fvm: address code review

Stebalien:


fvm.go
```diff
@@ -84,6 +95,10 @@ func CreateFVM(fvmVersion uint64, externs cgo.Externs, epoch abi.ChainEpoch, bas
 }
 
 func (f *FVM) ApplyMessage(msgBytes []byte, chainLen uint) (*ApplyRet, error) {
+	// NOTE: we need to call KeepAlive here (and below) because go doesn't guarantee that the
+	// receiver will live to the end of the function. If we don't do this, go _will_ garbage
+	// collecting the FVM, causing us to run the finalizer while we're in the middle of using
+	// the FVM.
```
dignifiedquire, line 101: For my  understanding, would this need go away if there were other go types in the FVM struct?

Stebalien: Only if we accessed them after our last use of the FVM in every function. But the keepalive is simply easier to reason about.

PR #251, opened 2022-03-28
fix identity

turuslan:
Fix regression from #249 ([fake.rs](https://github.com/filecoin-project/filecoin-ffi/pull/249/files#diff-c1ecb968cfe1f28e114eeb3b28aa1c484370f91199fc741482508b94b838ad31L38-R47)).

`fil_builtin_actors_bundle` car files use arbitrary string keys (identity hash) instead of hashes, but [`rust-multihash`](https://github.com/multiformats/rust-multihash/blob/master/src/hasher_impl.rs#L223) identity hasher can't "hash" such big values (actor wasm code).

without that fix rust panics when calling fvm functions

### DIFF
--- a/rust/src/fvm/blockstore/fake.rs
+++ b/rust/src/fvm/blockstore/fake.rs
@@ -6,6 +6,7 @@ use cid::{
     Cid,
 };
 use fvm_shared::blockstore::Blockstore;
+use fvm_shared::IDENTITY_HASH;
 
 use super::OverlayBlockstore;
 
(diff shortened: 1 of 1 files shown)


Stebalien (changes requested):
We've removed the "identity" feature from the multihash dependency in the filecoin-ffi dependency tree so rust-multihash shouldn't even have the identity feature.

turuslan:
This PR uses `IDENTITY_HASH` from #382, which was mentioned in #249 ([fake.rs](https://github.com/filecoin-project/filecoin-ffi/pull/249/files#diff-c1ecb968cfe1f28e114eeb3b28aa1c484370f91199fc741482508b94b838ad31L38-R47))

Stebalien:
That change was intentional. The bug is that we _shouldn't_ be using the identity feature. I'm currently fixing that issue.

Stebalien:
Closing in favor of #252.

PR #252, opened 2022-03-28
fix: completely remove the multihash identity feature

Stebalien:
And switch to rust edition 2021, with the version-2 feature resolver. This fixes the panic in the FVM constructor.

See https://github.com/filecoin-project/filecoin-ffi/pull/251.


Stebalien:
Note: we bump the edition/resolver to avoid having testing/build features "infect" all other build targets.

arajasek (approved):

PR #257, opened 2022-04-01
Update README.md

jennijuju:
given now we have built-in actor and fvm 😉


cryptonemo (approved):

cryptonemo:
> given now we have built-in actor and fvm wink

Definitely, thanks!

PR #258, opened 2022-04-04
Update fvm integration for FIP0031

vyzo:
Required for https://github.com/filecoin-project/lotus/pull/8429
Blocked on https://github.com/filecoin-project/ref-fvm/pull/448 and crate release

Note: Cargo points to my local dev env, which will make CI very angry. I will update once https://github.com/filecoin-project/ref-fvm/pull/448 merges and the crate is released.

### DIFF
--- a/fvm.go
+++ b/fvm.go
@@ -42,6 +42,7 @@ type FVMOpts struct {
 	BaseCircSupply abi.TokenAmount
 	NetworkVersion network.Version
 	StateBase      cid.Cid
+	Manifest       cid.Cid
 }
 
 // CreateFVM creates a new FVM instance.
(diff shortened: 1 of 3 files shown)


arajasek (approved):

arajasek (approved):

Stebalien:
I am _really_ confused. Passing the manifest into the FVM was, IIRC, a _hack_ to get this working for nv15. For nv16, the manifest should be linked into the state-tree via the upgrade. Why are we passing it here?

Stebalien (dismissed):
(blocked)

vyzo:
I think that will happen starting with nv17, where we can pass None.

The migrate has to already have happened to be able to do that. Passing it in allows us to test.

Stebalien:
> I think that will happen starting with nv17, where we can pass None.

NV16 _is_ the network version with the manifest linked into the state-tree. And yes, the migration has to pass to test. Or we need to link in the manifest when we create the genesis state, which we can and should do.

Stebalien:
We likely need to pass in the CAR in https://github.com/filecoin-project/lotus/blob/a709a0a6824797da416dbdf9de345ae55004fbf5/chain/gen/genesis/genesis.go#L160.

Stebalien:
Discussed synchronously. We're going to keep this _for testing_.

- For the real NV16, we'll set the manifest in the upgrade.
- For future test networks, we'll set it in genesis.

But having the ability to _override_ the manifest is still useful.

Stebalien (approved):

PR #262, opened 2022-04-08
[DO NOT MERGE] patch ref-fvm to support nv16

vyzo:
Patches in https://github.com/filecoin-project/ref-fvm/pull/470 to fvm@v0.5.1 for https://github.com/filecoin-project/lotus/pull/8429

This doesn't need to be merged, we should just propagate the new fvm version when the ref-fvm pr is merged instead.


raulk:
No merge necessary.

PR #263, opened 2022-04-11
feat: catch panics in the cgo externs

Stebalien:
Otherwise, we could panic across an FFI boundary, which is undefined behavior.

### DIFF
--- a/cgo/errors.go
+++ b/cgo/errors.go
@@ -5,10 +5,21 @@ package cgo
 // #cgo pkg-config: ${SRCDIR}/../filcrypto.pc
 // #include "../filcrypto.h"
 import "C"
+import (
+	"fmt"
+	"os"
+	"runtime/debug"
+)
 
 const (
 	ErrInvalidHandle   = C.ERR_INVALID_HANDLE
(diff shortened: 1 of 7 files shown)


cgo/errors.go
```diff
@@ -5,10 +5,21 @@ package cgo
...
+	"fmt"
+	"os"
+	"runtime/debug"
+)
 
 const (
 	ErrInvalidHandle   = C.ERR_INVALID_HANDLE
 	ErrNotFound        = C.ERR_NOT_FOUND
 	ErrIO              = C.ERR_IO
+	ErrPanic           = C.ERR_PANIC
```
magik6k, line 18: Supernit: Put after ErrInvalidArgument to match rust order

magik6k (approved):

magik6k (approved):

generated/libs.go
```diff
@@ -1,10 +0,0 @@
-package generated
```
dignifiedquire, line 1: This should not be deleted, strange things happen if this file is removed in my experience

PR #264, opened 2022-04-12
fvm M2 experiments

vyzo:
# DO NOT MERGE

See https://github.com/filecoin-project/ref-fvm/issues/444
Companion to https://github.com/filecoin-project/ref-fvm/pull/485 and https://github.com/filecoin-project/builtin-actors/pull/216

### DIFF
--- a/cgo/extern.go
+++ b/cgo/extern.go
@@ -6,6 +6,8 @@ typedef const uint8_t* buf_t;
 */
 import "C"
 import (
+	"fmt"
+	"os"
 	"unsafe"
 
 	"github.com/filecoin-project/go-address"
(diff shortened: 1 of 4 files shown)


cgo/errors.go
```diff
@@ -16,7 +16,7 @@ const (
 	ErrNotFound        = C.ERR_NOT_FOUND
 	ErrIO              = C.ERR_IO
 	ErrInvalidArgument = C.ERR_INVALID_ARGUMENT
-	ErrPanic           = C.ERR_PANIC
+	ErrPanic           = -5
```
raulk, line 19: Why did we need to change this?

rust/src/fvm/types.rs
```diff
@@ -59,7 +57,7 @@ impl Default for fil_FvmMachineExecuteResponse {
         fil_FvmMachineExecuteResponse {
             error_msg: ptr::null(),
             status_code: FCPResponseStatus::FCPNoError,
-            exit_code: ExitCode::Ok as u64,
+            exit_code: 0,
```
raulk, line 60: `ExitCode::OK.value() as u64`?

vyzo: that should work, sorry for the hack :)

PR #265, opened 2022-04-14
Support FVM exec traces

arajasek:


Stebalien:
@arajasek see #266

rust/src/fvm/machine.rs
```diff
@@ -127,18 +136,20 @@ pub unsafe extern "C" fn fil_create_fvm_machine(
...
 
+        let mut machine_context = network_config.for_epoch(chain_epoch, state_root);
+
+        machine_context
+            .set_base_fee(base_fee)
+            .set_circulating_supply(base_circ_supply);
+
+        if tracing {
+            machine_context.enable_tracing();
+        }
```
Stebalien, line 147: so... this is why taking a boolean can be slightly nicer.

rust/src/fvm/machine.rs
```diff
@@ -307,3 +334,135 @@ fn import_actors(
...
+            _ => {
+                return Err(anyhow!("expected ExecutionEvent of type Call"));
+            }
+        },
+        msg_receipt: Receipt {
+            exit_code: ExitCode::OK,
+            return_data: RawBytes::default(),
+            gas_used: 0,
+        },
+        error: "".to_string(),
```
Stebalien, line 373: `String::new()`

rust/src/fvm/machine.rs
```diff
@@ -207,6 +218,22 @@ pub unsafe extern "C" fn fil_fvm_machine_execute_message(
             }
         };
 
+        if !apply_ret.exec_trace.is_empty() {
+            let mut trace_iter = apply_ret.exec_trace.iter();
```
Stebalien, line 222:
```suggestion
            let mut trace_iter = apply_ret.exec_trace.into_iter();
```

will save you some clones later

rust/src/fvm/machine.rs
```diff
@@ -307,3 +334,135 @@ fn import_actors(
...
+struct LotusTrace {
+    pub msg: Message,
+    pub msg_receipt: Receipt,
+    pub error: String,
+    pub subcalls: Vec<LotusTrace>,
+}
+
+fn build_lotus_trace(
+    new_call: &ExecutionEvent,
+    trace_iter: &mut std::slice::Iter<ExecutionEvent>,
```
Stebalien, line 348:
```suggestion
    trace_iter: &mut impl Iterator<Item=ExecutionEvent>,
```

(makes your life easier).

Stebalien: (assumes you use `into_iter`)

Stebalien (approved):
Nits, otherwise lgtm.

Stebalien (approved):

rust/src/fvm/machine.rs
```diff
@@ -307,3 +333,148 @@ fn import_actors(
...
+        msg_receipt: Receipt {
+            exit_code: ExitCode::OK,
+            return_data: RawBytes::default(),
+            gas_used: 0,
+        },
+        error: String::new(),
+        subcalls: vec![],
+    };
+
+    while let Some(trace) = trace_iter.next() {
```
raulk, line 376: This pretty much repeats a bulk of the logic in the ref-fvm Executor. This layer should not be forming receipts, but _receiving_ already-formed receipts. Also, replicating the error number to exit code mapping table for top-level calls is a smell.

Stebalien: This was intentional to keep the FVM clean. Lotus expects Receipts, but receipts are for top-level messages, not sub-calls. So we convert to a format that lotus expects inside the FFI, with the goal of eventually fixing Lotus.

PR #269, opened 2022-04-22
Use new go-state-types state accessors

arajasek:
Needs https://github.com/filecoin-project/go-state-types/pull/36


ZenGround0 (approved):

raulk (approved):
The fact that we no longer depend on a specific (outdated) actors version is a sign that this change was needed, even before FVM.

PR #270, opened 2022-04-22
feat: propagate backtraces as errors

Stebalien:
This PR sets the ActorError to propagate backtrace information.


vyzo (approved):
at last!

PR #272, opened 2022-04-28
fix: allow zero-length slice allocs

Stebalien:
They're fine. We can't represent them as null, but rust will internally represent them as 0x1. We could just _assume_ that, but it's safer to call into rust and let it handle this case.

If we don't do this, some of the lotus tests fail.

### DIFF
--- a/cgo/helpers.go
+++ b/cgo/helpers.go
@@ -57,16 +57,13 @@ func AsSliceRefUint64(goBytes []uint64) SliceRefUint64 {
 	}
 }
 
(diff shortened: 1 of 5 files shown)


cryptonemo (approved):
Feels wrong ... but ok

dignifiedquire:
> If we don't do this, some of the lotus tests fail.

Can we add a todo, to fix them though?

Stebalien:
Why? I mean, allocating zero-length buffers isn't that unusual. I assume these tests are testing "fake" proofs. 

Basically, I see why a zero-length proof should fail. I don't see why a zero-length allocation should fail.

PR #273, opened 2022-04-28
fvm: Engine per network

magik6k:
Depends on https://github.com/filecoin-project/ref-fvm/pull/494

### DIFF
--- a/rust/src/fvm/machine.rs
+++ b/rust/src/fvm/machine.rs
@@ -1,4 +1,4 @@
-use std::convert::TryFrom;
+use std::convert::{TryFrom, TryInto};
 use std::sync::Mutex;
 
 use anyhow::{anyhow, bail};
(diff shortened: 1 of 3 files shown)


Stebalien (approved):

rust/src/fvm/machine.rs
```diff
@@ -31,7 +33,8 @@ pub type CgoExecutor = DefaultExecutor<
 >;
 
 lazy_static! {
-    static ref ENGINE: fvm::machine::Engine = fvm::machine::Engine::default();
+    static ref ENGINES: Mutex<BTreeMap<NetworkVersion, fvm::machine::Engine>> =
```
dignifiedquire, line 36: why BTreeMap, is the sorting relevant?

rust/src/fvm/machine.rs
```diff
@@ -147,8 +151,29 @@ pub unsafe extern "C" fn fil_create_fvm_machine(
         let blockstore = blockstore.finish();
 
         let externs = CgoExterns::new(externs_id);
+
+        let engine = match (|| -> anyhow::Result<Engine> {
+            let mut engines = ENGINES
+                .lock()
+                .map_err(|_| anyhow::Error::msg("engines lock is poisoned"))?;
```
dignifiedquire, line 158: `anyhow::anyhow!` is the more common way of doing this, especially including the error would be nice if you do this

dignifiedquire, line 158: in general there is no need to catch this, you can just unwrap it, given this indicates a previous panic, and we already catch panics

Stebalien (approved):
LGTM after release.

PR #279, opened 2022-05-16
feat: update FVM to v0.8.0

Stebalien:


rust/src/fvm/externs.rs
```diff
@@ -27,7 +26,7 @@ impl CgoExterns {
 impl Rand for CgoExterns {
     fn get_chain_randomness(
         &self,
-        pers: DomainSeparationTag,
+        pers: i64,
```
dignifiedquire, line 29: why this change?

Stebalien: This is mostly a change to enable M2, when users will want to specify their own. Basically, the actors care about these tags, but nobody else does.

PR #281, opened 2022-05-27
chore: upgrade the FVM to 1.0.0-rc2

Stebalien:
Also upgrades wasmtime to 0.37 (~20% perf improvement in some cases).


Stebalien:
> ~20% perf improvement in some cases

Not that this seems to matter for us.

arajasek (approved):

PR #287, opened 2022-06-09
fix: backport a fix for SSE feature detection

Stebalien:
Otherwise, wasmtime assumes sse4_1 support, and will fail on CPUs that only support SSE2. Unfortunately, I kind of doubt that we'll see a wasmtime release to fix this before we need to ship a release.


Stebalien:
I'm not a fan of patches in "production" code, but I'm not sure if we have a better option at the moment.

cryptonemo (approved):

Stebalien:
cc @lemmih. This change isn't consensus critical, but you may need to patch this crate on forest as well if you want to support older CPUs without SSE4.1 and SSE4.2 support.

PR #294, opened 2022-06-22
feat: update to latest v11.1.1 proofs

cryptonemo:


Stebalien (approved):
I would prefer stricter constraints, but this should fix the build (at least in lotus)

PR #303, opened 2022-07-07
lotus may crash: avoid rust vec use memory allocated by CGO's C.Bytes()

linguohua:
Hi, every one, currently CgoBlockstore::get() return a rust vector who use a buffer that allocated by CGO 's C.Bytes(), and it will crash when rust drop that buffer.

When build with rust's default allocator, lotus run as expected, but when use mimalloc_rust as rust global allocator, lotus will crash at some fvm relative functions.

I try to fix:

avoid rust vec to use memory allocated by CGO C.Bytes
add callback to cgo_blockstore_get to avoid memory copy
now lotus work correctly when use mimalloc_rust as rust's global allocator.

### DIFF
--- a/cgo/blockstore.go
+++ b/cgo/blockstore.go
@@ -11,6 +11,7 @@ import (
 /*
 #include <stdint.h>
 typedef const uint8_t* buf_t;
+extern void rust_vec_extend_func(const uint8_t* data, int32_t len, void* vec);
 */
 import "C"
 
@@ -24,7 +25,7 @@ func toCid(k C.buf_t, kLen C.int32_t) cid.Cid {
 }
 
 //export cgo_blockstore_get
-func cgo_blockstore_get(handle C.uint64_t, k C.buf_t, kLen C.int32_t, block **C.uint8_t, size *C.int32_t) (res C.int32_t) {
+func cgo_blockstore_get(handle C.uint64_t, k C.buf_t, kLen C.int32_t, v* C.void) (res C.int32_t) {
 	defer func() {
 		if rerr := recover(); rerr != nil {
 			logPanic(rerr)
@@ -39,8 +40,7 @@ func cgo_blockstore_get(handle C.uint64_t, k C.buf_t, kLen C.int32_t, block **C.
 	}
 
 	err := externs.View(ctx, c, func(data []byte) error {
-		*block = (C.buf_t)(C.CBytes(data))
-		*size = C.int32_t(len(data))
+		C.rust_vec_extend_func((*C.uint8_t)(&data[0]), C.int32_t(len(data)), unsafe.Pointer(v))
 		return nil
 	})
 
--- a/rust/src/fvm/blockstore/cgo.rs
+++ b/rust/src/fvm/blockstore/cgo.rs
@@ -1,4 +1,4 @@
-use std::ptr;
+//use std::ptr;
 
 use anyhow::{anyhow, Result};
 use cid::Cid;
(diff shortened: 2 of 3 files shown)


Stebalien:
At the moment, we don't support non-system allocators with the FFI, but you can pretty easily override the system allocator (see the [mimalloc](https://github.com/microsoft/mimalloc) documentation for details).

If you can provide sufficient motivation, I'm happy to revisit this PR if you can:

1. Demonstrate that this method doesn't have a significant performance impact over calling `C.CBytes`. I assume this is equivalent, but this is performance critical code.
2. Clean it up a bit. Instead of passing an empty vec from rust to go, use [AllocSliceBoxedUint8](https://github.com/filecoin-project/filecoin-ffi/blob/98eeb1eb8e45967c8ac8108a466819cc072f0292/cgo/helpers.go#L60), then return that.

(but I can't make any guarantees as this is performance and correctness critical code)

Stebalien:
Hm. So, thinking about this a bit, I can't see a way to reliably support non-system allocators. This code links to other rust libraries (e.g., OpenCL) which probably assume the system allocator as well.

Stebalien:
> What I mean is, in my opinion, we should not allocate memory in one language, than free it in another language.

This isn't a "go" allocation, it's a system allocation. `C.CBytes` uses the system allocator.

> Why currently the default rust allocator not crash lotus is because rust default allocator use the CRT(C Runtime) malloc/free, but rust not promise that, maybe someday it will use its own allocator that different from the CRT's.

For pure-rust applications, it makes no guarantees. For C/static libraries, it guarantees the system allocator for this very reason:

> Currently the default global allocator is unspecified. **Libraries, however, like cdylibs and staticlibs are guaranteed to use the [System](https://doc.rust-lang.org/std/alloc/struct.System.html) by default.**

This crate is a staticlib.

> Why we need to customise rust current default allocator?

You can still use any allocator you want, it just needs to be setup as your system allocator (through LD_PRELOAD, on linux).

Stebalien:
Again, I'm asking for:

1. Strong motivation for why using LD_PRELOAD is worse than patching this crate.
2. Evidence that this doesn't affect performance.

linguohua:
> Again, I'm asking for:
> 
> 1. Strong motivation for why using LD_PRELOAD is worse than patching this crate.
> 2. Evidence that this doesn't affect performance.

rust has clear way to specify global allocator:  https://doc.rust-lang.org/stable/std/alloc/trait.GlobalAlloc.html,  that is why we don't use LD_PRELOAD way.

linguohua:
> > What I mean is, in my opinion, we should not allocate memory in one language, than free it in another language.
> 
> This isn't a "go" allocation, it's a system allocation. `C.CBytes` uses the system allocator.
> 
> > Why currently the default rust allocator not crash lotus is because rust default allocator use the CRT(C Runtime) malloc/free, but rust not promise that, maybe someday it will use its own allocator that different from the CRT's.
> 
> For pure-rust applications, it makes no guarantees. For C/static libraries, it guarantees the system allocator for this very reason:
> 
> > Currently the default global allocator is unspecified. **Libraries, however, like cdylibs and staticlibs are guaranteed to use the [System](https://doc.rust-lang.org/std/alloc/struct.System.html) by default.**
> 
> This crate is a staticlib.
> 
> > Why we need to customise rust current default allocator?
> 
> You can still use any allocator you want, it just needs to be setup as your system allocator (through LD_PRELOAD, on linux).

I know  C.bytes() use the CRT malloc.
But I want to point out that, beforce filecoin-ffi 1.16, we use customise allocator for rust everything is ok, and then when we update to 1.16, it crash.  and the reason is just because we allocate memory use C.bytes() in go then free it in rust.
And from now on, every one can't use rust customise global allocator for lotus rust code. I think it is not a good thing.

Anyway, thanks for your conversation. I can close this pull request, and keep these code for myself use.

PR #311, opened 2022-07-22
fix #305: compile out rust-gpu-tools if both cuda and opencl are disabled

### LINKED ISSUE #305: Cannot build from source with FFI_USE_GPU=0
Even with FFI_USE_GPU=0 rust-gpu-tools are compiled and fail (rightfully) with:

error: At least one of the features `cuda` or `opencl` must be enabled.

I guess rust-gpu-tools should be a conditional dependency.

shawnrader:


rust/src/util/api.rs
```diff
@@ -67,10 +82,13 @@ pub fn init_log_fd(log_fd: libc::c_int) -> repr_c::Box<InitLogFdResponse> {
...
 #[cfg(test)]
 mod tests {
-
+    #[cfg(any(feature = "opencl", feature = "cuda"))]
     use crate::util::api::get_gpu_devices;
+
+    #[cfg(any(feature = "opencl", feature = "cuda"))]
     use crate::util::types::destroy_gpu_device_response;
 
+    #[cfg(any(feature = "opencl", feature = "cuda"))]
```
vmx, line 91: In case you don't want to repeat the feature flags, you can also do the imports directly in the test function. So it would look like:


```rust
    #[cfg(any(feature = "opencl", feature = "cuda"))]
    #[test]
    #[allow(clippy::needless_collect)]
    fn test_get_gpu_devices() {
        use crate::util::api::get_gpu_devices;
        use crate::util::types::destroy_gpu_device_response;

        let resp = get_gpu_devices();
        …
    }
```

rust/src/util/api.rs
```diff
@@ -32,16 +33,30 @@ pub fn init_log_with_file(file: File) -> Option<()> {
     }
 }
 
+// Serialize the GPU device names into a vector
```
vmx, line 36: This could be a doc comment (even if it's a private function) with three leading slashes.

rust/src/util/api.rs
```diff
@@ -32,16 +33,30 @@ pub fn init_log_with_file(file: File) -> Option<()> {
...
+        .collect();
+
+    devices
+}
+
+// Return empty vector for GPU devices if cuda and opencl are disabled
+#[cfg(not(any(feature = "opencl", feature = "cuda")))]
+fn get_gpu_devices_internal() -> Vec<slice_boxed<u8>> {
+    let v: Vec<slice_boxed<u8>> = Vec::new();
+    v
```
vmx, line 52: I should work to just put `Vec::new()` into the function body.

rust/src/util/api.rs
```diff
@@ -32,16 +33,30 @@ pub fn init_log_with_file(file: File) -> Option<()> {
...
+// Serialize the GPU device names into a vector
+#[cfg(any(feature = "opencl", feature = "cuda"))]
+fn get_gpu_devices_internal() -> Vec<slice_boxed<u8>> {
+    let devices = rust_gpu_tools::Device::all();
+    let devices: Vec<_> = devices
+        .into_iter()
+        .map(|d| d.name().into_bytes().into_boxed_slice().into())
+        .collect();
+
+    devices
```
vmx, line 45: The less code version without rebinding the variable would be:

```suggestion
    devices
        .into_iter()
        .map(|d| d.name().into_bytes().into_boxed_slice().into())
        .collect()
```

vmx (commented):
The CI failures seem unrelated to this change. Perhaps retriggering CI on master branch would make sense to see if it also happens there.

My inline comments are only about stylistic things, so feel free to ignore them. I just like take code review as a chance to give folks hints about idiomatic Rust/how I'd do things (which doesn't necessarily need to be better).

rust/src/util/api.rs
shawnrader: Originally I had this as a doc comment, but changed it since it was only called internally.  If there preference it to have it as a doc comment I can switch it back.

cryptonemo (approved):

rust/src/util/api.rs
vmx: My take on it is: if it really documents the function, why not making it a doc comment even if it's internal.

PR #314, opened 2022-08-04
feat: remove FVM actors-v7 bundle

Stebalien:
This was for testing the upgrade and is no longer necessary or useful. Removing this will shrink the build by a sizable amount.


arajasek (approved):
yes, YES!

PR #315, opened 2022-08-04
feat: update to latest proofs

cryptonemo:


cryptonemo:
Not urgent: This PR is broken, but signals that I'll need some FVM dependency update support here to the latest proofs @Stebalien @raulk @arajasek

cryptonemo:
Re: our recent discussion -- if it's not a good time to update proofs for latest work, let me know any ideas on how to move forward for lotus integration v1.17 (open to branching strategies, etc)

Stebalien:
We have a v1 branch in the FVM, so updating proofs there is pretty easy. Just tell me when and I'll do it.

cryptonemo:
> 

Perfect -- over time, this kind of thing should become easier as it becomes more clear what branch to target, etc!  @arajasek Can you assist with calling out the timing?  The intent was to have an FFI PR up in some form for the lotus code freeze next week.

jennijuju:
@cryptonemo lotus code freeze is tmr, so if we can get it in before 1500 EST it would be ideal

cryptonemo:
> @cryptonemo lotus code freeze is tmr, so if we can get it in before 1500 EST it would be ideal

Perfect, thanks!  @Stebalien Would you mind applying the dep updates in the appropriate FVM/v1 branch and push here? If you're otherwise busy, let me know and either @shawnrader or I can take a look.

jennijuju:
> We have a v1 branch in the FVM, so updating proofs there is pretty easy. 
@Stebalien this is not what we've discussed. I think the consensus was to land things (non-consensus breaking) in master, and lotus will test the ffi master on a monthly cadence.

Can we look into whats blocking the proof integration to master?

jennijuju:
> > We have a v1 branch in the FVM, so updating proofs there is pretty easy.
> 
> @Stebalien this is not what we've discussed. I think the consensus was to land things (non-consensus breaking) in master, and lotus will test the ffi master on a monthly cadence.
> 
> Can we look into whats blocking the proof integration to master?

cryptonemo:
At least the way I *think* it works from here is like this:

- FVM `v1` branch is updated to point to new proofs and deps
- FVM release is cut based on that `v1` branch
- This FFI PR is updated to point to the new `v1` based FVM crate
- CI passes and Lotus can integrate :tada:

Stebalien:
Updated to FVM 1.2 and bls-signatures 0.12.

shawnrader (approved):

PR #321, opened 2022-09-16
Update Cargo.lock dependencies for new merkletree v0.22.1 release

shawnrader:


cryptonemo (approved):
Without touching anything proofs, but having the new fallback to readonly store opens on read/write open failure, the reported issue should be resolved.  Let's see how CI does, then should be good for lotus integration/testing

PR #324, opened 2022-10-12
multifvm

vyzo:
This adds support for multiple fvm versions behind the same facade to the client, with dispatch based on the network version.
It allows us to host multiple network versions in the same client, including in development things, without blocking ourselves on backwards compatibility.

### DIFF
--- a/rust/src/fvm/engine.rs
+++ b/rust/src/fvm/engine.rs
@@ -272,9 +272,7 @@ impl AbstractMultiEngine for MultiEngine2 {
         externs: CgoExterns,
     ) -> InnerFvmMachine {
         let cfg = NetworkConfig2 {
-            network_version: unsafe {
-                std::mem::transmute::<NetworkVersion, NetworkVersion2>(cfg.network_version)
-            },
+            network_version: NetworkVersion2::try_from(cfg.network_version as u32).unwrap(),
             max_call_depth: cfg.max_call_depth,
             max_wasm_stack: cfg.max_wasm_stack,
             builtin_actors_override: cfg.builtin_actors_override,
--- a/rust/src/fvm/engine.rs
+++ b/rust/src/fvm/engine.rs
@@ -71,6 +71,12 @@ impl MultiEngineContainer {
     }
 }
 
+impl Default for MultiEngineContainer {
+    fn default() -> MultiEngineContainer {
+        MultiEngineContainer::new()
+    }
+}
+
 // fvm v3 implementation
 mod v3 {
     use cid::Cid;
--- a/rust/src/fvm/engine.rs
+++ b/rust/src/fvm/engine.rs
@@ -4,19 +4,43 @@ use std::sync::{Arc, Mutex};
 
 use anyhow::anyhow;
 use cid::Cid;
+
+use fvm2::call_manager::DefaultCallManager as DefaultCallManager2;
 use fvm3::call_manager::DefaultCallManager as DefaultCallManager3;
+
+use fvm2::executor::{
+    ApplyKind as ApplyKind2, ApplyRet as ApplyRet2, DefaultExecutor as DefaultExecutor2,
+    ThreadedExecutor as ThreadedExecutor2,
+};
 use fvm3::executor::{
-    ApplyKind as ApplyKind3, ApplyRet as ApplyRet3, DefaultExecutor as DefaultExecutor3,
-    ThreadedExecutor as ThreadedExecutor3,
+    ApplyKind, ApplyRet, DefaultExecutor as DefaultExecutor3, ThreadedExecutor as ThreadedExecutor3,
+};
+
+use fvm2::machine::{
+    DefaultMachine as DefaultMachine2, MachineContext as MachineContext2,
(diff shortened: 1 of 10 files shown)


rust/src/fvm/engine.rs
```diff
@@ -0,0 +1,239 @@
+use std::collections::hash_map::Entry;
+use std::collections::HashMap;
+use std::sync::{Arc, Mutex};
+
+use anyhow::anyhow;
+use cid::Cid;
+
+use fvm2::call_manager::DefaultCallManager as DefaultCallManager2;
```
Stebalien, line 8: Can we split this into an `v2` and an `v3`? That way we wouldn't need to rename all of our imports.
I.e.:

```rust
mod engine {
    mod v2 { ... }
    mod v3 { ... }
    pub use v2::new_executor as new_executor2;
    pub use v3::new_executor as new_executor3;
}
```

rust/src/fvm/engine.rs
```diff
@@ -0,0 +1,239 @@
...
+        raw_length: usize,
+    ) -> anyhow::Result<ApplyRet> {
+        use fvm2::executor::Executor;
+        let res = self.0.execute_message(
+            Message2 {
+                version: msg.version,
+                from: Address2::from_bytes(&msg.from.to_bytes()).unwrap(),
+                to: Address2::from_bytes(&msg.to.to_bytes()).unwrap(),
+                sequence: msg.sequence,
+                value: unsafe { std::mem::transmute::<TokenAmount3, TokenAmount2>(msg.value) },
```
Stebalien, line 107: can we avoid this? It won't really cost anything.

Stebalien: Yeah, let's get rid of the transmutes. We won't notice if something changes.

Stebalien (commented):
Approach LGTM. One comment about modules, one about transmutes.

Stebalien:
So, FUN TIMES! It looks like wasmtime is exporting some symbols, so we can't include it twice (in it's current incarnation, that is).

So... I think we need to:

1. Upgrade fvm v2 to wasmtime 1.0.
2. Test it _extensively_ on mainnet to make sure it's completely compatible.

https://github.com/filecoin-project/ref-fvm/issues/962

Stebalien:
Depends on https://github.com/filecoin-project/ref-fvm/pull/963.

rust/src/fvm/engine.rs
vyzo: removed all the transmutes; this was painful, but much better than have a random (maybe silent) failure.

Obligatory comment about rust: it is insane that the layout of structs and stuff is not deterministic. This is the biggest advantage of C, you can rely on things being the same and laid out the same.

vmx: Obligatory reply to rust comment :) If you want C semantics, you can annotate the structs with [`#[repr(C)]` ](https://doc.rust-lang.org/reference/type-layout.html#the-c-representation)`.

rust/src/fvm/engine.rs
vyzo: done; we don't even need to use the executors, it is handled internally.

vyzo:
ok, this is done; rebasing on master and marking as ready for review.

rust/src/fvm/engine.rs
vyzo: I dont necessarily want C repr, i jist want the compiler to be sane and predictable damnit!

Side note: i have somewhat warmed to that language of yours, but i still get fits from all those little things.

Stebalien (approved):
I am so, SO sorry. This looks like hell.

rust/src/fvm/engine.rs
```diff
@@ -0,0 +1,370 @@
...
+}
+
+impl Default for MultiEngineContainer {
+    fn default() -> MultiEngineContainer {
+        MultiEngineContainer::new()
+    }
+}
+
+// fvm v3 implementation
+mod v3 {
```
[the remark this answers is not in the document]
Stebalien: Eh, we can do that in a followup.

Stebalien:
Ok, so, this doesn't work on mainnet.

vyzo:
what happened?

Stebalien:
> what happened?

General "state mismatch" errors. Let me get you access to the box.

Stebalien:
Note: I've now tried FVM 2.1 and that's not the issue. Now I'm trying the few patches I just pushed, just in case.

Stebalien:
I've now checked with the patches I just pushed, and they don't help.

Stebalien:
I've confirmed that:

- Messages aren't just flat-out failing.
- The network config that we end up passing into the machine looks sane.

Looking at the diffs... I'm seeing _mostly_ state mismatches, but also some value mismatches. Strangely, I'm not just seeing random messages failing, so more debugging required...

Stebalien:
I found that we weren't invoking the threaded executor, so I fixed that. But that hasn't changed the situation.

Stebalien:
We will never mention this again.

vyzo:
ouch, that was not the right randomness.... sorry bout that!

PR #330, opened 2022-11-08
Build fvm3 from master without m2-native

vyzo:


vyzo:
closing in favour of release.

PR #338, opened 2022-11-18
feat: use the new unsafe.Slice method instead of casting arrays

Stebalien:
Previously, in cgo, we had to rely on casting pointers to fixed-sized arrays then slicing. But upstream has added an `unsafe.Slice` method, so we can remove this hack.

### DIFF
--- a/cgo/blockstore.go
+++ b/cgo/blockstore.go
@@ -93,14 +93,14 @@ func cgo_blockstore_put_many(handle C.uint64_t, lengths *C.int32_t, lengthsLen C
 		return ErrInvalidArgument
 	}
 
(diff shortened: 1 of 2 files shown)


cgo/blockstore.go
```diff
@@ -93,14 +93,14 @@ func cgo_blockstore_put_many(handle C.uint64_t, lengths *C.int32_t, lengthsLen C
 		return ErrInvalidArgument
 	}
 
-	lengthsGo := (*[MAX_LEN]C.int32_t)(unsafe.Pointer(lengths))[:lengthsLen:lengthsLen]
+	lengthsGo := unsafe.Slice(lengths, lengthsLen)
```
Stebalien, line 96: https://pkg.go.dev/unsafe#Slice

vyzo (approved):
that looks like an improvement; probably is just as unsafe of course :)

PR #340, opened 2022-11-29
Add aarch64 Linux build, test, and publish to CI

shawnrader:
- Rename circleci artifact tarbal names to differentiate x86_64 and aarch64

 - Add circleci build, test, and publish for aarch64

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,6 +30,25 @@ jobs:
       - run:
           command: make go-lint
 
+  build_and_test_aarch64_linux_cgo_bindings:
+    parameters:
+      run_leak_detector:
+        type: boolean
+        default: true
+    machine:
+      image: ubuntu-2004:current
+    resource_class: arm.large
+    working_directory: ~/go/src/github.com/filecoin-project/filecoin-ffi
+    steps:
+      - configure_environment_variables
+      - prepare
+      - build_project
+      - restore_parameter_cache
+      - obtain_filecoin_parameters
+      - save_parameter_cache
+      - run: cd rust && rustup target add wasm32-unknown-unknown
+      - run_tests
+
   build_and_test_linux_cgo_bindings:
     parameters:
       run_leak_detector:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -221,8 +245,8 @@ commands:
           command: |
             cd rust
 
-            TARBALL_PATH="/tmp/${CIRCLE_PROJECT_REPONAME}-$(uname)-standard.tar.gz"
-            RELEASE_NAME="${CIRCLE_PROJECT_REPONAME}-$(uname)-standard"
+            TARBALL_PATH="/tmp/${CIRCLE_PROJECT_REPONAME}-$(uname)-$(uname -m)-standard.tar.gz"
+            RELEASE_NAME="${CIRCLE_PROJECT_REPONAME}-$(uname)-$(uname -m)-standard"
```
Stebalien, line 249: This will probably require a change in lotus. @vmx any ideas?

shawnrader: It's possible to keep the name the same by splitting up the jobs in config.yml, but it is simpler and I think more descriptive to have architecture (aarch64, x86-64) in the name in the long run.

vmx: I really don't know much about the Lotus FFI integration. But to me it makes sense to have things named including their architecture if we plan to publish those as well. Thanks for flagging @Stebalien, we'll talk with the Lotus folks about it.

ianconsolata: Right now, the lotus build process relies mainly on the `install-filcrypto` script in the root of this project (https://github.com/filecoin-project/filecoin-ffi/blob/master/install-filcrypto#L79), so as long as that still works (or is updated to reference the new names) I think the lotus process should continue to function. Based on my reading of the script, I think it just takes the release name and appends `.tar.gz` which it looks like is what you're doing here as well, so I don't think it should require any change.

Once unzipped, are the libraries still called the same names?
```
filcrypto.h
libfilcrypto.a
filcrypto.pc
```

shawnrader: I just pushed a commit to update the release_name in install-filcrypto script to add the platform architecture to the tarball name.  Tested it out to verify it pulls down the binary and is able to build with pre-compiled source.

vmx (commented):
LGTM! But someone from the Lotus side should approve it.

ianconsolata (commented):
It looks like @jennijuju's PR testing this change passed (https://github.com/filecoin-project/lotus/pull/9768), though it didn't actually test building / releasing anything. I think this seems good enough for now. If not, we'll catch it on the next 1.19.0-rc release, and I can dig into it more if that breaks.

vmx (approved):

hammertoe:
A polite nudge on this. I'm trying to build Docker images for aarch64 Linux to allow developers to run dockerised instances of Lotus on M1s. If we can get this in then it save my build from having to build filecoin-ffi from scratch (and running out of disk space in the Github action runner in the process!). Thanks!

PR #344, opened 2022-12-02
chore: remove unnecessary dependencies

### LINKED ISSUE #342: Remove unused Rust dependencies
In the Cargo.toml there is a dependency on `bellperson`, but it's not used in the source code.

It would make sense to go over all dependencies to see which ones are actually used.

Stebalien:
fixes #342


vyzo (approved):

vmx (approved):
Ooooohhhh sweeeeet!

PR #345, opened 2022-12-07
chore: use stable rust version

Stebalien:
This is the rust version used by the actors.


Stebalien:
Gahhhhhhhhh: https://github.com/filecoin-project/bellperson/pull/269

Stebalien:
But someone else can handle that.

PR #348, opened 2022-12-18
FVM: Simplify Randomness externs, perform hashing on the Rust side

arajasek:
This is a refactor intended to ultimately move the hashing portion of the randomness calculation into the FVM itself.

### DIFF
--- a/rust/src/fvm/externs.rs
+++ b/rust/src/fvm/externs.rs
@@ -1,5 +1,8 @@
 use anyhow::{anyhow, Context};
+use byteorder::{BigEndian, WriteBytesExt};
+use std::io::Write;
 
+use blake2b_simd::Params;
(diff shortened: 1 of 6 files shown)


rust/src/fvm/externs.rs
```diff
@@ -113,6 +102,22 @@ impl Rand2 for CgoExterns {
     }
 }
 
+fn draw_randomness(
```
Stebalien, line 105: Can we have one or two "fixture" tests? I.e.:

1. In lotus, explicitly generate some randomness (e.g., with a unit test).
2. Hard-code those in a test in _this_ file, and make sure this function generates the same outputs.

Stebalien (approved):

arajasek:
Closing, stale

PR #360, opened 2023-02-07
chore: update FVM to 3.0.0-alpha.23

Stebalien:
This updates the FVM and switches gas to u64.


rust/src/fvm/engine.rs
```diff
@@ -273,8 +277,8 @@ mod v2 {
                         ret.over_estimation_burn.atto().clone(),
                     ),
                     refund: TokenAmount::from_atto(ret.refund.atto().clone()),
-                    gas_refund: ret.gas_refund,
-                    gas_burned: ret.gas_burned,
+                    gas_refund: ret.gas_refund.try_into().context("negative gas refund")?,
+                    gas_burned: ret.gas_burned.try_into().context("negative gas burned")?,
```
Stebalien, line 281: Ditto, the FVM will never return negative values here.

rust/src/fvm/engine.rs
```diff
@@ -318,12 +322,26 @@ mod v2 {
                         .into_iter()
                         .filter_map(|tr| match tr {
                             ExecutionEvent2::GasCharge(charge) => {
+                                // We set the gas for "negative" charges to 0. This isn't correct,
+                                // but it won't matter in practice (the sum of the gas charges in
+                                // the trace aren't guaranteed to equal the total gas charge
+                                // anyways).
```
Stebalien, line 328: @arajasek has said that this is fine. We only use tracing for benchmarking/debugging gas anyways.

arajasek (approved):

PR #362, opened 2023-02-08
Add support for h-nv20

Stebalien:


rust/src/fvm/engine.rs
```diff
@@ -180,10 +184,302 @@ mod v3 {
...
+                            emitter: e.emitter,
+                            event: ActorEvent {
+                                entries: e
+                                    .event
+                                    .entries
+                                    .into_iter()
+                                    .map(|e| Entry {
+                                        flags: Flags::from_bits(e.flags.bits().into()).unwrap(),
+                                        key: e.key,
+                                        codec: 0,
```
raulk, line 419: 👍

rust/src/fvm/engine.rs
```diff
@@ -180,10 +184,302 @@ mod v3 {
...
+                                    },
+                                    Cause3h1::Fatal {
+                                        error_msg,
+                                        backtrace,
+                                    } => Cause::Fatal {
+                                        error_msg,
+                                        backtrace,
+                                    },
+                                }),
+                            })
```
raulk, line 343: Do we have actual data structure changes here? If not, is not possible to transmute the structs?

raulk (approved):

raulk:
Would've preferred a merge commit to bring the branch up to speed with master, and then the patch, but it's fine this way too.

rust/src/fvm/engine.rs
Stebalien: Transmuting without specified representation is always UB. Rust can,e.g., decide to re-order enum tags based on usage (it probably won't, but I don't want to play that game).

PR #366, opened 2023-02-12
fix: set chain id correctly for hyperspace.

raulk:
We were not setting the chain ID correctly for "v3h1" 😢


Stebalien:
Ah, sorry. I adapted the "conversion" code from v2, not v3 (because v3 doesn't _do_ any conversion).

PR #367, opened 2023-02-13
Adds a version filter for release tags to perform the publishing

cryptonemo:
This adds a bash script filter to only tag and publish releases if a tag of the form 'v*' exists, which supports standard versioning such as 'vx.y.z'


cryptonemo:
The 2 options for this are breaking up the CI config into separate build/package/publish steps (currently all bundled since those happen via script), or adding simple logic like this (in script) that accomplishes the same thing.

rust/scripts/publish-release.sh
```diff
@@ -18,6 +18,14 @@ main() {
     local __release_file=$1
     local __release_name=$2
     local __release_tag="${CIRCLE_SHA1:0:16}"
+    local __release_tag_name="${CIRCLE_TAG:-default}"
+
+    if [[ ! -z ${__release_tag_name} && ${__release_tag_name} =~ ^v.*$ ]]; then
```
vmx, line 23: Perhaps it makes sense to add a comment like "only publish if a release tag starting with `v` was pushed.

vmx (approved):
One comment, but it looks good.

PR #368, opened 2023-02-13
fix: fvm: remove all unwraps

Stebalien:
We were catching panics here, but the unwraps ended up hiding the actual error message because rust backtraces are disabled by default.


arajasek (approved):
Seems reasonable

PR #372, opened 2023-02-14
backport: fix: fvm: remove all unwraps

arajasek:
We were catching panics here, but the unwraps ended up hiding the actual error message because rust backtraces are disabled by default.


arajasek:
Nvm, let's just merge

PR #379, opened 2023-02-28
feat: plumb codec through lotus trace

Stebalien:
This replaces the trace types with FVM specific ones. It:

- Adds codecs to params/return values.
- Removes unset fields like nonce, gas, etc.


Stebalien:
This will require some extensive changes in lotus, so I'll tackle those first.

arajasek (approved):
LGTM, thank you. Can we drop this TODO about the backtrace? I _think_ we're sending that info up to Lotus? https://github.com/filecoin-project/filecoin-ffi/blob/8e6542762efa3dbb45f2a4082e8e0b6c4c7a3113/rust/src/fvm/machine.rs#L306

PR #381, opened 2023-03-02
Enable CUDA as default GPU backend if available

cryptonemo:
If CUDA is found in the environment, it will be enabled unless `FFI_USE_OPENCL=1` is set.  If CUDA is not found by default or if building on Darwin, OpenCL will be used instead.  If `FFI_NO_GPU=1`, no GPU support will be used.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -283,6 +283,7 @@ commands:
             echo 'export PATH="/usr/local/go/bin:${HOME}/.cargo/bin:${PATH}:${HOME}/go/bin:${HOME}/.bin"' >> $BASH_ENV
             echo 'export RUST_LOG=info' >> $BASH_ENV
             echo 'export CIRCLE_ARTIFACTS="/tmp"' >> $BASH_ENV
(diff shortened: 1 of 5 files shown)


vmx (commented):
I just want to mention that if you compile with CUDA and OpenCL support, whatever is available will be selected at runtime. So we could also compile with both and still CUDA would kind of be the default.

Though I don't know if that would be better. In this current PR you can easily compile with CUDA only and OpenCL only. That would be harder if we also allow compiling with both enabled (which you usually wouldn't need as you know if you've CUDA or OpenCL).

vmx (commented):
I think you should also make `cuda` the default in the `Cargo.toml` file to match the expectations. Also in case someone is not using those shell scripts for building.

vmx (approved):
I think it would be good if also someone from the Lotus team reviews it even if it's a "OK for me, it doesn't really matter for us"-review.

Stebalien (approved):
As long as this will continue to work without requiring CUDA, I don't see why not?

Has anyone been testing the CUDA backend in production?

jennijuju (approved):

PR #383, opened 2023-03-09
chore: update proofs

Stebalien:
This updates the proofs and the FVM, getting rid of a bunch of deprecated dependencies in the process.


cryptonemo (approved):
Looks good, thanks @Stebalien!

Issue #343, opened 2022-12-02
Is nightly Rust still needed?


vmx:
Currently the Rust parts are built with `nightly-2022-05-09`. On my local dev machine, I can't even get that version installed. Hence I propose checking if nightly is really needed or whether we can switch to a stable release.

Stebalien:
We needed it for a proofs thing for a while, but I think that's gone.

Stebalien:
Fixed!

PR #384, opened 2023-03-09
chore: switch to rust stable

Stebalien:
And update the toolchain. Now that we've updated the proofs libraries, we're no longer tied to nightly.


vmx:
I'm looking into this right now.

vmx:
Sadly, we're still tied to nightly https://github.com/filecoin-project/filecoin-ffi/issues/343.

Stebalien:
Ah, damn. I thought we had fixed all of those. Thanks!

PR #385, opened 2023-03-10
chore: use Rust stable on CI

vmx:
While doing that, the whole CI got a bit of a clean up. E.g. use stock circle CI docker images instead of custom ones. Also downscale the resources classes that were used. It now takes less than 25% of the credits.

Also update the code to fix any warnings caused by using a newer Rust version.

This PR supersedes https://github.com/filecoin-project/filecoin-ffi/pull/384. It will pass CI once there is a proofs release that includes https://github.com/filecoin-project/rust-fil-proofs/pull/1679.


Stebalien (approved):

Stebalien:
Apparently we need to test on stable but build on nightly.

Stebalien:
Ah. I think we can just go ahead and do that. Thanks for investigating!

vmx:
Finally! :tada:

PR #387, opened 2023-03-14
feat: implement FIP-0061

cryptonemo:
~~Ready to go for review/testing, but does require proofs releases to be referenced, rather than git `master` branches.~~

Good to go! :rocket:


Stebalien:
I've updated the FVM and rebased on master.

arajasek (approved):

PR #388, opened 2023-03-14
feat: make the FVM concurrency configurable and set the default to 4

Stebalien:
Previously, we were limited to 1 message at a time.


arajasek (approved):
LGTM, we could consider sanity checking the value (non-zero, have some max, etc.)

Stebalien:
> LGTM, we could consider sanity checking the value (non-zero, have some max, etc.)

Good point.

arajasek (approved):

scotthconner:
It doesn't look like it actually defaults to 4? as we changed the lazy static to new_env() versus new()?

PR #390, opened 2023-03-29
FVM3: Support nv19 and 20

arajasek:
As in the title.


Stebalien (approved):

Stebalien:
Er, wait, don't we need to update/release ref-fvm first?

cryptonemo (approved):

arajasek:
> Er, wait, don't we need to update/release ref-fvm first?

We _also_ need to do that. I'm just looking to get things working first (with the FVM and actors patched for now), but I'm landing smaller things like this change.

PR #392, opened 2023-04-05
feat: implement FIP-0061

arajasek:
- Add necessary CGO constants
- Refactor `toFilPublicReplicaInfo`:
  - For Winning PoSt, maintain previous behaviour
  - For Window PoSt, we no longer have a function mapping SealProof to PoSt proof, so simply use the PoSt type supplied. Note that this method is only used by miners locally to sanity check their proofs, so weakening security is acceptable here (you can only "attack" yourself).


cryptonemo (approved):
Looks good, thanks!

PR #398, opened 2023-04-23
chore: deps: update to FVM 3.3 crates

arajasek:


raulk (approved):
This works for mainnet, but for calibnet we may need to do something different depending on what the solution is: nv20 (in which case we might need a major...), or no nv20 (in which case this might be fine).

PR #402, opened 2023-05-19
feat: add various new APIs for Lotus

vmx:
Those new APIs are about:

 - decoding a range of an updated empty sector
 - generating TreeRLast
 - generating TreeC
 - running SDR

### DIFF
--- a/rust/src/proofs/api.rs
+++ b/rust/src/proofs/api.rs
@@ -150,17 +150,20 @@ fn seal_pre_commit_phase1(
     })
 }
 
-/// TODO: document
+/// Runs the SDR process the same way as it would during PreCommit Phase 1.
+///
+/// The `output_dir` is the directory where the layer labels are stored. The `replica_id` is needed
+/// to make sure the output is unique.
 #[ffi_export]
 fn generate_sdr(
     registered_proof: RegisteredSealProof,
-    cache_path: c_slice::Ref<u8>,
+    output_dir: c_slice::Ref<u8>,
     replica_id: &[u8; 32],
 ) -> repr_c::Box<GenerateSdrResponse> {
     catch_panic_response("generate_sdr", || {
         seal::sdr(
             registered_proof.into(),
-            as_path_buf(&cache_path)?,
+            as_path_buf(&output_dir)?,
             (*replica_id).into(),
         )?;
 
@@ -192,7 +195,10 @@ fn seal_pre_commit_phase2(
     })
 }
 
-/// TODO: document
+/// Generates a TreeRLast the same way as during PreCommit Phase 2 and return the CommRLast.
+///
+/// The `replica_path` points to the sealed file. `output_dir` is where the TreeRLast should be
+/// stored. It's a directory as it may consist of serveral files.
 #[ffi_export]
 fn generate_tree_r_last(
     registered_proof: RegisteredSealProof,
@@ -210,7 +216,14 @@ fn generate_tree_r_last(
     })
 }
 
-/// TODO: document
+/// Generates a TreeC the same way as during PreCommit Phase 2 and returns the CommC.
+///
+/// The `input_dir` is the directory where the label layers are stored, which were constructed
+/// during the SDR process (PreCommit Phase 1). No other data is needed.
+/// The `output_dir` is the directory where the resulting TreeC tree is stored (it may be split
+/// into several files).
+/// The `input_dir` and `output_dir` may point to the same directory. Usually that's the "cache
+/// directory".
 #[ffi_export]
 fn generate_tree_c(
     registered_proof: RegisteredSealProof,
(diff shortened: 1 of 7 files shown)


rust/src/proofs/api.rs
```diff
@@ -150,6 +150,24 @@ fn seal_pre_commit_phase1(
     })
 }
 
+/// TODO: document
+#[ffi_export]
+fn generate_sdr(
+    registered_proof: RegisteredSealProof,
+    cache_path: c_slice::Ref<u8>,
+    replica_id: &[u8; 32],
```
magik6k, line 158: Are you aware of any Go code utils for generating replica_id? (Or at least some docs / easy to understand rust code so that we can implement thin in Go)

rust/src/proofs/api.rs
```diff
@@ -174,6 +192,42 @@ fn seal_pre_commit_phase2(
...
+        Ok(comm_r_last)
+    })
+}
+
+/// TODO: document
+#[ffi_export]
+fn generate_tree_c(
+    registered_proof: RegisteredSealProof,
+    input_dir: c_slice::Ref<u8>,
+    output_dir: c_slice::Ref<u8>,
```
magik6k, line 218: Are both of those the "cache directory"? I'm assuming it's fine if we set both to the same path.

rust/src/proofs/api.rs
```diff
@@ -150,6 +150,24 @@ fn seal_pre_commit_phase1(
     })
 }
 
+/// TODO: document
+#[ffi_export]
+fn generate_sdr(
+    registered_proof: RegisteredSealProof,
+    cache_path: c_slice::Ref<u8>,
```
magik6k, line 157: I'm assuming this only generates layers, and we're only interested in the sc-02-data-layer-11.dat.

Separately, does that keep all the layers on disk, or just the last (two) one? Could we add a flag to switch between the two modes if that's easy enough?

rust/src/proofs/api.rs
```diff
@@ -174,6 +192,42 @@ fn seal_pre_commit_phase2(
...
+
+        Ok(comm_r_last)
+    })
+}
+
+/// TODO: document
+#[ffi_export]
+fn generate_tree_c(
+    registered_proof: RegisteredSealProof,
+    input_dir: c_slice::Ref<u8>,
```
magik6k, line 217: What files exactly need to be in the input dir? `sc-02-data-layer*` + `?_aux`?

rust/src/proofs/api.rs
```diff
@@ -830,6 +884,34 @@ fn empty_sector_update_decode_from(
...
+#[ffi_export]
+unsafe fn empty_sector_update_decode_from_range(
+    registered_proof: RegisteredUpdateProof,
+    comm_d: &[u8; 32],
+    comm_r: &[u8; 32],
+    input_fd: libc::c_int,
+    sector_key_fd: libc::c_int,
+    output_fd: libc::c_int,
+    nodes_offset: u64,
+    num_nodes: u64,
```
magik6k, line 897: Just sanity checking this is in fr32 elements (so basically padded bytes * 32)?

magik6k (approved):

magik6k:
(meant to comment)

rust/src/proofs/api.rs
vmx: I wasn't sure if Lotus has that already, it seems it doesn't. It should be easy to implement. The code is at https://github.com/filecoin-project/rust-fil-proofs/blob/95e8c982540b14077d58ee9ddc076f7602b4805d/storage-proofs-porep/src/stacked/vanilla/params.rs#L758-L775. It really is just a normal SHA256 hash, nothing special.

If the inputs don't make sense (we have slightly different naming in our code base), let me know.

rust/src/proofs/api.rs
vmx: Yes, in your case they would both point to the same "cache directory". I just wanted to propagate the flexibility to have different ones upwards, in case it's needed in the future.

rust/src/proofs/api.rs
vmx: Currently it's keep all layers around. I'm not sure how much `rust-fil-proofs` currently is into deleting files. Currently SDR is also internally kind of this single big operation API wise. So I'd prefer if it would be OK if we don't remove files while it's running. So would a "delete_all_layers_except_the_last_one_flag" help? Or, if it's done at the end of the whole SDR, would it be done on the Go side then?

rust/src/proofs/api.rs
vmx: I'll add proper docs.

rust/src/proofs/api.rs
vmx: Yes, it's the number of elements, though it;s padded_bytes **divided** by 32.

cryptonemo:
This PR works as-is, but the Cargo deps aren't rebased against master.  It's not clear why to be yet, but when rebased against master things break.  I've made an alternate PR that has things properly rebased, but it requires disabling CUDA (which of course isn't ideal, but may be a starting spot for proper resolution).

That PR is here for reference: https://github.com/filecoin-project/filecoin-ffi/pull/410

cryptonemo:
Related: https://github.com/filecoin-project/rust-gpu-tools/issues/84

A fix and a new patch release of rust-gpu-tools is required to re-enable CUDA support

cryptonemo:
@magik6k Properly rebased with CUDA re-enabled.  I think this is what you want.

cryptonemo (approved):

magik6k (approved):

PR #405, opened 2023-06-02
fix: build for CUDA and other fixes

vmx:
This change makes CI build things on CUDA (which is the default). It also fixes various other things, like making the `go fmt` job actually work. It also uses smaller instances where it didn't increase the time too much.

---

@magik6k is it a problem to test with the most recent Go 1.18 instead of 1.17.9? I'm asking as this would make the CI significantly simpler as the CircleCI convenience images for Go >= 1.18 already use Ubuntu 22.04, older versions use 20.04.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,12 +6,14 @@ orbs:
 executors:
   golang:
     docker:
-      - image: cimg/go:1.17.9
+      - image: cimg/go:1.18
     resource_class: small
-  rust:
-    docker:
-      - image: cimg/rust:1.67
-    resource_class: small
+    environment:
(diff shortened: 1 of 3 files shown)


cryptonemo (approved):
Looks good to me, but please get approval from someone on Lotus and/or FVM before merging

Stebalien:
> @magik6k is it a problem to test with the most recent Go 1.18 instead of 1.17.9? I'm asking as this would make the CI significantly simpler as the CircleCI convenience images for Go >= 1.18 already use Ubuntu 22.04, older versions use 20.04.

The latest stable release of lotus relies has a minimum go version of 1.18, so we absolutely should update. The latest feature release uses 1.20.

PR #400, opened 2023-05-09
Synthetic PoRep

cryptonemo:

PR #407, opened 2023-06-19
feat: add FFI_USE_CUDA_SUPRASEAL env variable

vmx:
Make it possible to build the FFI with SupraSeal enabled, which should lead to better performance during C2.


vmx:
I close that one, as I've rebased that commit directly to the synthetic porep branch at https://github.com/filecoin-project/filecoin-ffi/pull/400.

PR #409, opened 2023-06-29
chore: update fvm for wasmtime v10 and cid v0.10

Stebalien:
This also unifies both FVM versions on the same ipld encoding crate.

This version also introduces wasm instance "oversubscription" so we can allocate fewer wasm instances (reserve less address space) while specifying a high concurrency factor.


Stebalien:
Ok, I've tested it with fvm v2 epochs (around 2004818) and everything is working.

arajasek (approved):

PR #412, opened 2023-08-14
FVM: Simplify Randomness externs, perform hashing on the Rust side

arajasek:
Integrates FVM change: https://github.com/filecoin-project/ref-fvm/pull/1842.

FVM3 no longer specifies a DST and entropy when requesting randomness.
FVM3 still does, so we perform the hashing required to "draw" the randomness here. We could alternatively refactor FVM2 to have the same change, feedback welcome.


rust/src/fvm/externs.rs
```diff
@@ -42,15 +45,8 @@ impl Rand3 for CgoExterns {
...
-                self.handle,
-                pers,
-                round,
-                entropy.as_ptr(),
-                entropy.len() as i32,
-                &mut buf,
-            ) {
-                0 => Ok(buf),
+            match cgo_extern_get_chain_randomness(self.handle, round, &mut buf) {
+                0 => draw_randomness(&buf, pers, round, entropy),
```
Stebalien, line 49: We're doing this in the FVM now, right?

arajasek:
Syncing mainnet well

Stebalien (approved):

PR #421, opened 2023-09-28
fix: install aarch64-apple-darwin toolchain on x86

vmx:
We always create universal binaries when compiling on x86. This means that the aarch64-apple-darwin also needs to be installed. Install that toolchain automatically if it doesn't exist yet.

This change has also been tested for the case when the toolchain already exists. In that case it just reports:

    info: component 'rust-std' for target 'aarch64-apple-darwin' is up to date

Thanks @lanzafame for reporting this issue.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -103,7 +103,6 @@ jobs:
       - prepare:
           linux: false
           darwin: true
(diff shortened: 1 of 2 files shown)


cryptonemo (approved):
Looks good, but like a broken record, I'll point to `build.sh` again :sweat_smile:

Issue #420, opened 2023-09-28
Releasing MacOS prebuilts will break in June 2024


vmx:
[CircleCI is sunsetting all Intel MacOS machines in 2024](https://discuss.circleci.com/t/macos-intel-support-deprecation-in-january-2024/48718). Currently we use such a machine to create universal binaries.

We should switch to an ARM based machine and cross-compile to Intel from there (if that's possible, I don't know much about MacOS).

**Update**: the sunsetting was postponed till June 2024.

galargh:
https://github.com/filecoin-project/filecoin-ffi/blob/62960ca63d5fb1649ad95a947182175bbd5f11e8/install-filcrypto#L162-L164 suggests that it might not be possible.

Is dropping darwin/amd64 support an option?

Alternatively, let me know if you wanted us to look into setting up self-hosted Apple Intel runners.

vmx:
A possible fix is at https://github.com/filecoin-project/filecoin-ffi/pull/424/.

PR #424, opened 2023-10-02
ci: switch macos builds to apple silicon and upgrade xcode

galargh:
This PR migrated the macos builds in the repo to Apple Silicon executors since the Apple Intel ones are getting deprecated by January 2024. See https://discuss.circleci.com/t/macos-intel-support-deprecation-in-january-2024/48718.

FYI, the builds currently fail with:
```
# runtime/cgo
ld: warning: directory not found for option '-L/usr/lib/aarch64-linux-gnu/stubs'
# net
ld: warning: directory not found for option '-L/usr/lib/aarch64-linux-gnu/stubs'
# github.com/filecoin-project/filecoin-ffi/cgo
ld: warning: directory not found for option '-L/usr/lib/aarch64-linux-gnu/stubs'
ld: library not found for -lhwloc
clang: error: linker command failed with exit code 1 (use -v to see invocation)
FAIL    github.com/filecoin-project/filecoin-ffi [build failed]
```

I assume it's going to be easier to figure out for maintainers.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -93,9 +93,9 @@ jobs:
 
   build_darwin_cgo_bindings:
     macos:
-      xcode: "12.5.1"
+      xcode: "13.4.1"
     working_directory: ~/go/src/github.com/filecoin-project/filecoin-ffi
-    resource_class: macos.x86.medium.gen2
+    resource_class: macos.m1.medium.gen1
     steps:
       - configure_environment_variables:
           linux: false
@@ -316,6 +316,7 @@ commands:
                 name: Add a few more environment variables
                 command: |
                   echo 'export PATH="${HOME}/.cargo/bin:${HOME}/.bin:${PATH}"' >> $BASH_ENV
+                  echo 'export LIBRARY_PATH=/opt/homebrew/lib' >> $BASH_ENV
   obtain_filecoin_parameters:
     steps:
       - run: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -104,7 +104,6 @@ jobs:
           linux: false
           darwin: true
       - run: cd rust && cargo fetch
-      - run: cd rust && cargo install cargo-lipo
       - build_project
       - compile_tests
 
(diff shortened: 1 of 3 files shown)


vmx:
This is even more difficult. We currently create universal binaries on MacOS for x86 and aarch64. We need to check if it also works to create x86 version on the M1. I've opened https://github.com/filecoin-project/filecoin-ffi/issues/420 for tracking that.

vmx:
I've fixed the original error, it was just about setting the `LIBRARY_PATH`. Though this should *not* be merged as building the universal binary on MacOS probably does *not* work yet.

vmx:
This PR needs thorough review/testing by someone from the Lotus team. It should build the correct release artifacts, but I haven't really tested it. I'm neither that familiar with Lotus, nor do I own a Mac.

rvagg:
@galargh @vmx I hope you don't mind that I rebased and force pushed this branch to get it updated.

Tried this successfully on my m2:

```
$ file libfilcrypto.a
./libfilcrypto.a: Mach-O universal binary with 2 architectures: [x86_64:current ar archive] [arm64]
./libfilcrypto.a (for architecture x86_64):	current ar archive
./libfilcrypto.a (for architecture arm64):	current ar archive
```

Though compiling lotus:

```
ld: warning: dylib (/opt/homebrew/lib/libhwloc.dylib) was built for newer macOS version (14.0) than being linked (13.3)
```

I have xcode 14 locally, I guess brew built hwloc for me, but the Circle config has `13.4.1`. I'm just not sure how 13.3 got in here if I also built the FFI locally, I don't see an xcode or macos-version specifier in here anywhere. Maybe it's a rust thing? Hopefully the difference between 13.4 and 13.3 isn't going to make this a common warning.

But lotus still runs, and syncs the chain from a snapshot.

`make unittests` works, although I get the `ld` warning each time it runs something that touches ffi, it's pretty noisy. I get some unittest failures in storage/sealer involving `lstat /dev/fd/4: bad file descriptor` and also mentioning `DEBUG	ffiwrapper	ffiwrapper/sealer_cgo.go:218	AddPiece: cannot close pieceData reader *nullreader.NullReader because it is not an io.Closer`, which is sus, but I get the same errors if I run it from a normal build. So I don't think it's related.

I have an Intel mac in the house I could test the universal binary on just to be sure, but I don't have access to it at the moment, I'll try later in the week.

vmx:
@rvagg I'm happy that you have a look. I have almost zero experience with MacOS (and no local machine to test on).

vmx (approved):

PR #425, opened 2023-10-04
feat: add FFI_USE_CUDA_SUPRASEAL env variable

vmx:
Make it possible to build the FFI with SupraSeal enabled, which should lead to better performance during C2.

This was accidentally removed from the PR at https://github.com/filecoin-project/filecoin-ffi/pull/418.

PR #426, opened 2023-10-04
feat: add FFI_USE_CUDA_SUPRASEAL env variable (#425)

vmx:
Make it possible to build the FFI with SupraSeal enabled, which should lead to better performance during C2.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -108,6 +108,19 @@ jobs:
       - run: cd rust && cargo install cargo-lipo
       - build_project
       - compile_tests
+
+  # SupraSeal pulls in a C++ code base, make sure everything compiles properly.
+  build_linux_supraseal:
+    executor: golang
+    resource_class: medium
+    working_directory: ~/go/src/github.com/filecoin-project/filecoin-ffi
+    steps:
+      - configure_environment_variables
+      - prepare
+      - run:
+          name: Build project with `FFI_USE_CUDA_SUPRASEAL=1`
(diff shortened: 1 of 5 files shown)


magik6k (approved):

Stebalien:
Wait for @arajasek before merging.

arajasek:
I don't think we want this in 1.24.0. It's landed in `master`, so it'll go out in the next release?

vmx:
We pushed hard to get this into 1.24. It was just accidentally not merged.

arajasek:
^What @Stebalien said. Basically, 1.24.0 should be _exactly_ what's in https://github.com/filecoin-project/filecoin-ffi/commits/release/lotus-v1.23.4, plus anything _needed_ for SynthPoRep / nv21.

vmx:
> This _will_ be in 1.25 which will be released at the same time.

Just to make sure I understand this correctly: whether this change is in 1.24 or 1.25, people will be able to enable this feature *on the same day*. The only difference is that they need to use 1.25 instead of 1.24, correct?

arajasek:
> > This _will_ be in 1.25 which will be released at the same time.
> 
> Just to make sure I understand this correctly: whether this change is in 1.24 or 1.25, people will be able to enable this feature _on the same day_. The only difference is that they need to use 1.25 instead of 1.24, correct?

Yes, that is correct.

vmx:
> Yes, that is correct.

Cool, I just wanted to make sure that feature gets in the hands of people as soon as possible, which is the case (I don't mind which version that is). Hence I'm closing his PR.

Issue #431, opened 2023-10-30
CI fails on ARM due to cgocheck > 1 mode is no longer supported at runtime


vmx:
Currently master if failing on CI: https://app.circleci.com/pipelines/github/filecoin-project/filecoin-ffi/3025/workflows/6477ffa8-3057-4d33-82f5-cdc665d478e1/jobs/18154

due to:

```
fatal error: cgocheck > 1 mode is no longer supported at runtime. Use GOEXPERIMENT=cgocheck2 at build time instead.

runtime stack:
runtime.throw({0x629cd8?, 0xa?})
        runtime/panic.go:1077 +0x40 fp=0xffffe98f7810 sp=0xffffe98f77e0 pc=0x48c30
runtime.parsegodebug({0x4000000020?, 0x7?}, 0x0)
        runtime/runtime1.go:498 +0x240 fp=0xffffe98f78b0 sp=0xffffe98f7810 pc=0x5b7f0
runtime.parsedebugvars()
        runtime/runtime1.go:406 +0x11c fp=0xffffe98f78f0 sp=0xffffe98f78b0 pc=0x5b3fc
runtime.schedinit()
        runtime/proc.go:744 +0x15c fp=0xffffe98f7940 sp=0xffffe98f78f0 pc=0x4c62c
runtime.rt0_go()
        runtime/asm_arm64.s:86 +0xa4 fp=0xffffe98f7970 sp=0xffffe98f7940 pc=0x7cca4

Exited with code exit status 2
```

Note that it only fails on ARM.

It seems to be triggered by these lines:

https://github.com/filecoin-project/filecoin-ffi/blob/441fa8e61189dc32c2960c1f8d8ba56269f20366/.circleci/config.yml#L354-L357

snadrus:
This error is from a Go 1.21 runtime. 
[My fix](https://github.com/filecoin-project/filecoin-ffi/pull/432) is failing because it's running on a Go 1.17 runtime. 

Your CI link avoids the [Go 1.17 install step](https://app.circleci.com/pipelines/github/filecoin-project/filecoin-ffi/3027/workflows/36284f68-9d15-4a03-aaea-000142551c91/jobs/18173/parallel-runs/0/steps/0-104). 

The proper command for Go 1.21+ is in branch: fix431
Otherwise the line is correct for Go 1.20-

[...]

PR #434, opened 2023-10-31
feat: add option to not require a t_aux file to be present

vmx:
The `FFI_USE_FIXED_ROWS_TO_DISCARD` variable will make the proofs ignore the `FIL_PROOFS_ROWS_TO_DISCARD` setting, hence not relying on a `t_aux` file to be present. If enabled, then no `t_aux` file will be written by the proofs and a hard-coded value for `FIL_PROOFS_ROWS_TO_DISCARD` will be used.

---

I make this into draft mode as CI currently isn't passing due to https://github.com/filecoin-project/filecoin-ffi/issues/431.

### DIFF
--- a/install-filcrypto
+++ b/install-filcrypto
@@ -208,11 +208,19 @@ build_from_source() {
         use_multicore_sdr=""
     fi
 
+    # By default the number or rows to discard of the TreeRLast can be set via
+    # `FIL_PROOFS_ROWS_TO_DISCARD`. When SupraSeal PC2 is used, then this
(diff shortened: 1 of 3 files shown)


cryptonemo (approved):
Enables latest proofs version

rjan90:
As we now have confirmed that SupraSeal PC2 works with https://github.com/filecoin-project/lotus/pull/11185#issuecomment-1814960653 this latest proofs-version I think we can merge this and bubble it up to Lotus.

PR #436, opened 2023-11-03
fix: remove ClearLayerData call

vmx:
The `ClearLayerData` FFI call was accidentally introduced with the Synthetic PoRep. The call does under the hood exactly what `ClearCache` is doing. It was already removed from Lotus in [PR 11352], so that we can remove it here as well (and as a next steps from the Rust side). This reduces the API surface, which is generally a good idea.

[PR 11352]: https://github.com/filecoin-project/lotus/pull/11352


vmx:
I've removed it from the Go side manually. If any of those files are generated somehow, that generation should be run instead.

Stebalien (approved):

Stebalien:
All the go code is hand written at the moment.

PR #445, opened 2023-12-21
refactor: clean up deps

vmx:
The `byteorder` crate isn't actually used.

Also remove the direct dependency on the `fr32` crate, which was used for testing only anway. As `fr32` is part of the proofs, this change should make upgrades easier, as now the FFI only depends on `filecoin-proofs-api`, which is the public API for proofs.


Stebalien (approved):
:heart:

PR #446, opened 2024-01-25
Update the FVM to v4.1

Stebalien:
- Supports nv22.
- Cleanups & refactors.
- Changes the trace format a little to include the state-root of every actor when invoked.

### DIFF
--- a/rust/src/fvm/engine.rs
+++ b/rust/src/fvm/engine.rs
@@ -74,7 +74,7 @@ impl TryFrom<u32> for EngineVersion {
         match value {
             16 | 17 => Ok(EngineVersion::V1),
             18 | 19 | 20 => Ok(EngineVersion::V2),
-            21 => Ok(EngineVersion::V3),
+            21 | 22 => Ok(EngineVersion::V3),
             _ => return Err(anyhow!("network version not supported")),
         }
     }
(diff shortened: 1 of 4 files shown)


rust/src/fvm/engine.rs
```diff
@@ -427,9 +427,6 @@ mod v3 {
                                         .unwrap_or(ErrorNumber::AssertionFailed),
                                 )))
                             }
-                            ExecutionEvent3::InvokeActor(cid) => {
-                                Some(ExecutionEvent::InvokeActor(cid))
```
Stebalien, line 431: I'm removing this information from v3 traces. It's only used by the ETH JSON-RPC API, which is moderately broken at the moment.

rust/src/fvm/machine.rs
```diff
@@ -384,7 +386,12 @@ pub struct TraceMessage {
     pub codec: u64,
     pub gas_limit: u64,
     pub read_only: bool,
-    pub code_cid: Cid,
+}
+
+#[derive(Serialize_tuple, Deserialize_tuple, Debug, PartialEq, Eq, Clone)]
+pub struct TraceActor {
```
Stebalien, line 392: This will require lotus changes. We introduced this change so we could determine the state of an actor _at_ the point where we invoke it (even if it never actually gets committed to the chain state, e.g., due to a revert).

The idea was to use this in the Eth JSON-RPC API.

arajasek (approved):
LGTM

PR #447, opened 2024-01-31
chore: update FVM to v4.1.2 and v3.9.0

rjan90:
Prepering a PR updating FVM to the new FVM-release (v4.1.2) that is coming, due to https://github.com/filecoin-project/ref-fvm/pull/1978

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -1165,13 +1165,13 @@ dependencies = [
  "filecoin-proofs-api",
  "filepath",
  "fvm 2.7.0",
- "fvm 3.8.0",
- "fvm 4.1.1",
+ "fvm 3.9.0",
+ "fvm 4.1.2",
  "fvm_ipld_blockstore",
  "fvm_ipld_encoding",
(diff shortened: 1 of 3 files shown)


Stebalien:
NOTE: I'll also be releasing a new version in the v3 line.

rjan90:
Ack! I will wait for the new version in the v3 line version to land and update that in this PR as well.

Stebalien:
@rjan90 this should be good to go. Can you give my code a review?

Stebalien:
(may be we should have just done this as two PRs...)

rjan90 (commented):
I do not have approval rights, but the changes for updating the versions looks good to me.

When it comes to the changes in `/fvm/engine.rs`, those are a little bit outside my knowledge area. But they do seem correct to me from the standpoint that we are now changing `ExecutionEvent::InvokeActor` to include the actor's ID and state-object (codeCID, balance, current state root, delegated address, and sequence/nonce)

Stebalien:
I tagged a release (need it in my lotus tracing work).

PR #455, opened 2024-05-07
chore: update Rust version on CI to 1.73

vmx:
Use the same rust version on CI as in the rust-toolchain.toml file.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -152,7 +152,7 @@ jobs:
       - publish_darwin_release
   rustfmt:
     docker:
-      - image: cimg/rust:1.67
+      - image: cimg/rust:1.73
     resource_class: small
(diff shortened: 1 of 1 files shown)


rvagg (approved):
should we maybe bump the version of both a bit further? ref-fvm is at 1.78 and builtin-actors is at 1.76
looks like rust-fil-proofs is down at 1.70.0 though

vmx:
> should we maybe bump the version of both a bit further?

That's a good question. For me the important part was to have CI in sync with the `rust-toolchain.toml`. So I\m merging this one.

In case we should have a higher version here, I think it's a separate PR bumping both. Often new Rust versions lead to new Clippy lints, so sometimes also code changes are needed in case the version is bumped.

PR #453, opened 2024-04-23
feat: add support for Non-interactive PoRep

vmx:
This commit adds support for ni-porep.

Also Rust >= 1.70 is needed due to the dependency on `home` v0.5.9 and fix the corresponding new Clippy warnings.

---

This is only the Rust side of things, the Go parts are missing.

PR #458, opened 2024-06-12
Ni-PoRep ref-fvm update

rvagg:
I split this out of https://github.com/filecoin-project/filecoin-ffi/pull/456; but it depends on releases from https://github.com/filecoin-project/ref-fvm/pull/2010 before it can be merged.

PR #456, opened 2024-05-27
Ni-PoRep Go/cgo additions

rvagg:
~Some additional changes on top of a rebased https://github.com/filecoin-project/filecoin-ffi/pull/453, while I try and wire things up for integration testing.~

~This has local linking for ref-fvm for https://github.com/filecoin-project/ref-fvm/pull/1999 on top of master for https://github.com/filecoin-project/ref-fvm/pull/2009.~

~It also has local linking for rust-filecoin-proofs-api for https://github.com/filecoin-project/rust-filecoin-proofs-api/pull/103~

--

Additions on top of https://github.com/filecoin-project/filecoin-ffi/pull/453, but this depends on https://github.com/filecoin-project/go-state-types/pull/269 being merged.

Note that his also includes `seal_commit_phase2_circuit_proofs` in `rust/src/proofs/api.rs`, that could go into #453.

PR #459, opened 2024-06-13
feat: add support for Non-interactive PoRep

vmx:
This commit adds support for ni-porep.

It's a combination of the indivdual PRs #453, #456 and #458. Thanks everyone involved working on those.

The changes are for the Go side (actors and FVM) as well as the Rust side (proofs API).

Changes on the CI are:

 - Makes it match the Rust version in the rust-toolchain.toml
 - Newer Go version (1.21) on all platforms
 - Larger instance for the testsas the Ni-PoRep synthesis phase takes more resources.

Closes #453, #456, #458.

---

I made a new PR containing all changes to make things easier to follow and review.

What is still missing is a new `rust-fil-proofs` release containing https://github.com/filecoin-project/rust-fil-proofs/pull/1758. To keep things simple, it'll probably be a patch release, so the only thing that we'll need will be updating the `Cargo.lock` of the `filecoin-ffi`.

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -63,3 +63,8 @@ c-headers = ["safer-ffi/headers"]
 # This feature enables a fixed number of discarded rows for TreeR. The `FIL_PROOFS_ROWS_TO_DISCARD`
 # setting is ignored, no `TemporaryAux` file will be written.
 fixed-rows-to-discard = ["filecoin-proofs-api/fixed-rows-to-discard"]
+
+[patch.crates-io]
+filecoin-proofs = { git = "https://github.com/filecoin-project/rust-fil-proofs", branch = "fip90-and-test-updates" }
+filecoin-proofs-api = { git = "https://github.com/filecoin-project/rust-filecoin-proofs-api", branch = "use-pub-more" }
+storage-proofs-core = { git = "https://github.com/filecoin-project/rust-fil-proofs", branch="fip90-and-test-updates" }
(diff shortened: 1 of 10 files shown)


rjan90 (approved):
I have done a review of the things that I have knowledge about, so it would be great to have someone else 👀 and ✅ on this as well. But:

- The constants in `cgo/const.go` looks correct.
- It now uses the GST `v0.14.0-rc1` which is the first GST release candidate for nv23.
- fvm4, fvm4_shared points to the latests versions.

rvagg (approved):
very excellent, let's do it!

Stebalien:
I pushed one more commit to update FVM v2 & v3 to remove the old proofs code from the build tree.

Stebalien (approved):

rust/Cargo.toml
```diff
@@ -55,11 +60,33 @@ tempfile = "3.0.8"
...
+  "fvm3/opencl",
+  "fvm4/opencl",
+]
 multicore-sdr = ["filecoin-proofs-api/multicore-sdr"]
 c-headers = ["safer-ffi/headers"]
 # This feature enables a fixed number of discarded rows for TreeR. The `FIL_PROOFS_ROWS_TO_DISCARD`
 # setting is ignored, no `TemporaryAux` file will be written.
 fixed-rows-to-discard = ["filecoin-proofs-api/fixed-rows-to-discard"]
+
+[patch.crates-io]
```
cryptonemo, line 89: Requires a new proofs release (of course, but just noting)

cryptonemo (approved):

rust/Cargo.toml
cryptonemo: Updated

Issue #463, opened 2024-06-28
Migrate CI from CircleCI to GitHub Actions


BigLep:
This is the Filecoin-FFI equivalent of https://github.com/filecoin-project/lotus/issues/11734

TODO: fill in more details

PR #464, opened 2024-07-05
ci: migrate ci setup from circle ci to github actions

galargh:
In this PR, we migrate the CI from CircleCI to GitHub Actions.

The important difference in the GitHub Actions setup is that we're running the Linux ARM64 builds on self-hosted GitHub Actions runners.

Code style checks, builds, and tests have been tested in CI - see https://github.com/filecoin-project/filecoin-ffi/actions/runs/9857064865

Publishing will only be verified after the merge. 

Before the merge, we have to stop building in CircleCI and change the Required workflows setting.

Once we verify the new workflow works as expected, we're going to remove the CircleCI configuration altogether.

As requested, the GitHub Actions release workflow supports releasing from an arbitrary historical refs using the workflow dispatch trigger with ref input.

### DIFF
--- a/.github/actions/configure-environment/action.yml
+++ b/.github/actions/configure-environment/action.yml
@@ -9,13 +9,13 @@ runs:
         echo 'GO111MODULE=on' >> $GITHUB_ENV
         echo 'RUST_LOG=info' >> $GITHUB_ENV
       shell: bash
-    - if: runner.os == 'Linux' && runner.arch == 'x86_64'
+    - if: runner.os == 'Linux' && runner.arch == 'ARM64'
       run: |
         # Make sure CUDA is found on aarch64
         echo 'LD_LIBRARY_PATH="/usr/lib/aarch64-linux-gnu/stubs:'"${LD_LIBRARY_PATH}"'"' >> $GITHUB_ENV
         echo 'LIBRARY_PATH="/usr/lib/aarch64-linux-gnu/stubs:'"${LIBRARY_PATH}"'"' >> $GITHUB_ENV
       shell: bash
-    - if: runner.os == 'MacOS'
+    - if: runner.os == 'macOS'
       run: |
         echo 'export LIBRARY_PATH=/opt/homebrew/lib' >> $GITHUB_ENV
       shell: bash
@@ -27,7 +27,7 @@ runs:
         sudo update-alternatives --install /usr/bin/c++ c++ /usr/bin/g++-10 10
         sudo update-alternatives --set c++ /usr/bin/g++-10
       shell: bash
-    - if: runner.os == 'MacOS'
+    - if: runner.os == 'macOS'
       run: |
         HOMEBREW_NO_AUTO_UPDATE=1 brew install pkg-config md5sha1sum jq hwloc || brew link --overwrite python@2
       shell: bash
--- a/.github/actions/configure-environment/action.yml
+++ b/.github/actions/configure-environment/action.yml
@@ -30,7 +30,7 @@ runs:
         echo "${HOME}/.cargo/bin" >> $GITHUB_PATH
         echo "${HOME}/.bin" >> $GITHUB_PATH
       shell: bash
-    - if: runner.os == 'Linux' && runner.arch == 'x64'
+    - if: runner.os == 'Linux' && runner.arch == 'ARM64'
       run: |
         echo "LD_LIBRARY_PATH=/usr/lib/aarch64-linux-gnu/stubs:${LD_LIBRARY_PATH}" >> $GITHUB_ENV
         echo "LIBRARY_PATH=/usr/lib/aarch64-linux-gnu/stubs:${LIBRARY_PATH}" >> $GITHUB_ENV
--- a/.github/actions/configure-environment/action.yml
+++ b/.github/actions/configure-environment/action.yml
@@ -16,8 +16,6 @@ runs:
         echo "${HOME}/.cargo/bin" >> $GITHUB_PATH
         echo "${HOME}/.bin" >> $GITHUB_PATH
       shell: bash
-    - if: runner.os == 'Linux' && runner.arch == 'ARM64'
-      uses: Jimver/cuda-toolkit@9b295696791d75d658d8de64c4a85097ad8abeaf # v0.2.16
     - if: runner.os == 'macOS'
       run: |
         ls "/Applications" | grep "Xcode_"
@@ -36,6 +34,17 @@ runs:
     - if: runner.os == 'macOS'
       run: |
         HOMEBREW_NO_AUTO_UPDATE=1 brew install pkg-config md5sha1sum jq hwloc libmx || brew link --overwrite python@2
+      shell: bash
+    - if: runner.os == 'Linux' && runner.arch == 'ARM64'
+      run: |
+        wget https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2204/sbsa/cuda-ubuntu2204.pin
+        sudo mv cuda-ubuntu2204.pin /etc/apt/preferences.d/cuda-repository-pin-600
+        wget https://developer.download.nvidia.com/compute/cuda/12.5.1/local_installers/cuda-repo-ubuntu2204-12-5-local_12.5.1-555.42.06-1_arm64.deb
+        sudo dpkg -i cuda-repo-ubuntu2204-12-5-local_12.5.1-555.42.06-1_arm64.deb
+        sudo cp /var/cuda-repo-ubuntu2204-12-5-local/cuda-*-keyring.gpg /usr/share/keyrings/
+        sudo apt-get update
+        sudo apt-get -y install cuda-toolkit-12-5
+        echo "LD_LIBRARY_PATH=/usr/local/cuda-12.5/lib64:/usr/local/cuda-12.5/lib64/stubs:$LD_LIBRARY_PATH" >> $GITHUB_ENV
       shell: bash
     - uses: dtolnay/rust-toolchain@21dc36fb71dd22e3317045c0c31a3f4249868b17
       with:
--- a/.github/actions/configure-environment/action.yml
+++ b/.github/actions/configure-environment/action.yml
@@ -1,0 +1,40 @@
+name: Configure Environment Variables
+description: Configure environment variables for Filecoin FFI
+
+runs:
+  using: 'composite'
+  steps:
+    - run: |
+        echo 'FIL_PROOFS_PARAMETER_CACHE="${GITHUB_WORKSPACE}/filecoin-proof-parameters/"' >> $GITHUB_ENV
+        echo 'GO111MODULE=on' >> $GITHUB_ENV
+        echo 'GOPATH="${GITHUB_WORKSPACE}/go"' >> $GITHUB_ENV
+        echo 'PATH="/usr/local/go/bin:${GITHUB_WORKSPACE}/.cargo/bin:${PATH}:${GITHUB_WORKSPACE}/go/bin:${GITHUB_WORKSPACE}/.bin"' >> $GITHUB_ENV
+        echo 'RUST_LOG=info' >> $GITHUB_ENV
+        echo 'CIRCLE_ARTIFACTS="/tmp"' >> $GITHUB_ENV
+        # Make sure CUDA is found on aarch64
+        echo 'LD_LIBRARY_PATH="/usr/lib/aarch64-linux-gnu/stubs:${LD_LIBRARY_PATH}"' >> $GITHUB_ENV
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,221 @@
...
+      - if: runner.os == 'macOS'
+        name: Build project and tests, but don't actually run the tests (used to verify that build/link works with Darwin)
+        run: GOEXPERIMENT=cgocheck2 RUST_LOG=info go test -run=^$
+  supraseal:
+    name: Build with CUDA supraseal
+    runs-on: ubuntu-latest
+    # # Is it OK if we operate in the default working directory?
+    # defaults:
+    #   run:
+    #     working-directory: go/src/github.com/filecoin-project/filecoin-ffi
```
laurentsenta, line 145: dead code?

bls_test.go
```diff
@@ -121,12 +121,10 @@ func BenchmarkBLSVerifyBatch(b *testing.B) {
 func benchmarkBLSVerifyBatchSize(size int) func(b *testing.B) {
 	return func(b *testing.B) {
 		var digests []Digest
-		var msgs []Message
 		var sigs []Signature
 		var pubks []PublicKey
 		for i := 0; i < size; i++ {
 			msg := Message(fmt.Sprintf("cats cats cats cats %d %d %d dogs", i, i, i))
-			msgs = append(msgs, msg)
```
laurentsenta, line 129: Is this expected? I guess that breaks some linting rules?

laurentsenta (approved):

.github/workflows/ci.yml
galargh: Yes, deleting.

bls_test.go
galargh: Yeah, the linting rules started failing. The issue made sense to me so I fixed it. I suspect we're using newer version of some linting tool or something like that.

.github/actions/configure-environment/action.yml
```diff
@@ -0,0 +1,61 @@
...
+    - if: runner.os == 'Linux' && runner.arch == 'ARM64'
+      run: |
+        wget -q https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2204/sbsa/cuda-ubuntu2204.pin
+        sudo mv cuda-ubuntu2204.pin /etc/apt/preferences.d/cuda-repository-pin-600
+        wget -q https://developer.download.nvidia.com/compute/cuda/12.5.1/local_installers/cuda-repo-ubuntu2204-12-5-local_12.5.1-555.42.06-1_arm64.deb
+        sudo dpkg -i cuda-repo-ubuntu2204-12-5-local_12.5.1-555.42.06-1_arm64.deb
+        sudo cp /var/cuda-repo-ubuntu2204-12-5-local/cuda-*-keyring.gpg /usr/share/keyrings/
+        sudo apt-get update
+        sudo apt-get -y install cuda-toolkit-12-5
+        sudo mkdir -p /usr/lib/aarch64-linux-gnu/stubs
```
rvagg, line 16: This is all new for GHA, is the arm64 image available to us different enough that there's no cuda available already?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,213 @@
...
+      - if: github.event.inputs.ref != ''
+        uses: actions/checkout@v4
+        with:
+          submodules: recursive
+          ref: ${{ github.event.inputs.ref }}
+      - name: Build project with `FFI_USE_CUDA_SUPRASEAL=1`
+        run: FFI_BUILD_FROM_SOURCE=1 FFI_USE_CUDA_SUPRASEAL=1 make
+  publish:
+    needs: [check, cgo-bindings, supraseal]
+    if: (github.event_name == 'push' && startsWith(github.ref, 'refs/tags/')) || github.event.inputs.publish == 'true'
```
rvagg, line 152: Is there no nicer way to specify "only run on `v.*` tag?"? Both this and its inverse above are fine, the intent is hidden under layers of abstraction which add to the general incomprehensibility of GHA config language.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,213 @@
...
+      - name: Run go fmt
+        # `! go fmt ./... 2>&1 | read"` doesn't work, this one does, thanks
+        # https://carsonip.me/posts/go-fmt-and-ci/
+        run: |
+          output=$(go fmt ./...)
+          echo "${output}"
+          test -z "${output}"
+      - name: Run various linters
+        run: |
+          go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.56.2
```
rvagg, line 79:
```suggestion
          go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.59.0
```

How about we jump up to the same version we've pinned lotus to. I don't think this introduces new failures, at least from my local testing.

bls_test.go
rvagg: https://hub.docker.com/layers/cimg/go/1.21.2-node/images/sha256-59b3fff1c4745e9195ea35eef2cfbee4cb0b562616869e304b581644020ad4b8 - golangci-lint v1.54.2 vs v1.56.2

but, I don't see any version of golangci-lint failing on this file. staticcheck does but I don't see it being used here?

rvagg: btw I'm not objecting to this fix, just interested in why this is showing up for this diff

rvagg:
Looks pretty close to circle config. Not at all clearer though! I worry a little about future maintainability of complex GHA configs like this, but I'm not sure what can be done in lieu of a better config language for GHA.

bls_test.go
galargh: This was the failing run before I added this change - https://github.com/filecoin-project/filecoin-ffi/actions/runs/9805936223/job/27076653921

It does come from `staticcheck` as you pointed out. I wonder why we weren't running it in CircleCI.

galargh: It is listed in the linters list there, too. Honestly, I'm not sure why it wasn't flagging this issue.

.github/workflows/ci.yml
galargh: I'm afraid that's the most straighforward way of making this check at the moment unless we wanted to split the workflow into multiple files.

.github/workflows/ci.yml
galargh: Sure! Let's give it a go.

BigLep:
For the record, this. closes https://github.com/filecoin-project/filecoin-ffi/issues/463

.github/actions/configure-environment/action.yml
rvagg: I'd like to hear @vmx on this; it's his area, cuda and nvidia driver versions for me are just a source of frustration rather than something I know a whole lot about!

bls_test.go
rvagg: `make go-lint` only runs golangci-lint. So how about we add staticcheck install to this workflow and to the Makefile too so we have it properly covered

rvagg:
@vmx should be back this week and hopefully he can do a quick review on the relevant bits of this, particularly commenting on the cuda bits in the discussion above, and then we should be good to :ship:

.github/actions/configure-environment/action.yml
vmx: For development I recommend installing CUDA drivers from the NVIDIA website as the distro ones often don't work well (for whatever reason). Though for CI, I think using the distro packages could work (KISS). If it works for Linux, it should also work for ARM (I hope).

We should support old enough versions and for filecoin-ffi it's more of a sanity check. So whatever the CUDA version is, as long as the tests pass, we are good.

bls_test.go
galargh: It is running in GHA - that's why I had to add the fix - so I wonder how much time we want to spend investigating this.

.github/actions/configure-environment/action.yml
```diff
@@ -0,0 +1,61 @@
...
+        sudo update-alternatives --set c++ /usr/bin/g++-10
+        # Check if we need to install cuda-toolkit-12-5
+      shell: bash
+    - if: runner.os == 'macOS'
+      run: |
+        HOMEBREW_NO_AUTO_UPDATE=1 brew install pkg-config md5sha1sum jq hwloc
+      shell: bash
+    - uses: dtolnay/rust-toolchain@21dc36fb71dd22e3317045c0c31a3f4249868b17
+      with:
+        toolchain: 1.73
```
vmx, line 58: A note on hard-coding things. It would be better if the version would be taken from the https://github.com/filecoin-project/filecoin-ffi/blob/d041c9ab85e68cfb57daf5fbaedeeee0f72cb5ad/rust/rust-toolchain.toml file. Something similar to https://github.com/multiformats/rust-multihash/blob/b321afc11e874c08735671ebda4d8e7fcc38744c/.github/workflows/build.yml#L86-L93 may be possible. Though this could be in a separate PR.

.github/actions/configure-environment/action.yml
galargh: Got it; I'd move on with installing it on linux arm during the run first, and once we add A version of it to the base image, we'll come back here and remove that step.

.github/actions/configure-environment/action.yml
galargh: Created an issue for this - https://github.com/filecoin-project/filecoin-ffi/issues/470

.github/actions/configure-environment/action.yml
```diff
@@ -0,0 +1,61 @@
...
+        wget -q https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2204/sbsa/cuda-ubuntu2204.pin
+        sudo mv cuda-ubuntu2204.pin /etc/apt/preferences.d/cuda-repository-pin-600
+        wget -q https://developer.download.nvidia.com/compute/cuda/12.5.1/local_installers/cuda-repo-ubuntu2204-12-5-local_12.5.1-555.42.06-1_arm64.deb
+        sudo dpkg -i cuda-repo-ubuntu2204-12-5-local_12.5.1-555.42.06-1_arm64.deb
+        sudo cp /var/cuda-repo-ubuntu2204-12-5-local/cuda-*-keyring.gpg /usr/share/keyrings/
+        sudo apt-get update
+        sudo apt-get -y install cuda-toolkit-12-5
+        sudo mkdir -p /usr/lib/aarch64-linux-gnu/stubs
+        sudo ln -s /usr/local/cuda-12.5/lib64/stubs/libcuda.so /usr/lib/aarch64-linux-gnu/stubs/libcuda.so.1
+        sudo ln -s /usr/local/cuda-12.5/lib64/stubs/libcuda.so /usr/lib/aarch64-linux-gnu/stubs/libcuda.so
```
vmx, line 18: I'm not sure if that's really needed. It looks a bit like a direct port from the CircleCI config. I'd expect the files to be accessible through `/usr/local/cuda/lib64/stubs/libcuda.so` (not that there's no specific cuda version in the path). And then later below, it should be enough to put the  `LD_LIBRARY_PATH` to `usr/local/cuda/lib64/stubs/` (if that's needed at all, perhaps it's already found automatically).

vmx (approved):
I've left comments, but generally it looks good and as long it's green I think it's good to be merged. Hence I'm approving it.

.github/actions/configure-environment/action.yml
galargh: I've tried the suggestion in https://github.com/filecoin-project/filecoin-ffi/pull/472, but it doesn't seem to cut it, unfortunately.

BigLep:
@galargh : anything blocking from getting this merged?

BigLep:
@rvagg : per Slack, @galargh believes this is good to merge.  Can you please handle?

galargh:
Thanks for merging. I've now disabled CircleCI in this repo. Please ping me when you prepare to create a new tag here, we'll verify if publishing from GitHub Actions works as expected.

rvagg:
https://github.com/filecoin-project/filecoin-ffi/pull/476 getting rid of circle config so it stops showing up in PRs; I assume this is the appropriate next step here.

> Please ping me when you prepare to create a new tag here

Not sure how much we have queued that'll need it, maybe there's a new ref-fvm upgrade coming down the pipe but I don't think it's imminent.

rvagg:
Of course we could just make a new tag, I don't think there'd be a problem with that now that we've got Lotus 1.28.0 done, a tag just for the sake of testing CI wouldn't be a bad thing I think. @rjan90 what do you think?

rjan90:
> Of course we could just make a new tag, I don't think there'd be a problem with that now that we've got Lotus 1.28.0 done, a tag just for the sake of testing CI wouldn't be a bad thing I think. @rjan90 what do you think?

Agree that we can just create a tag for testing CI, but let us schedule it after the Mainnet upgrade on `2024-08-06`, just to prevent any user confusion around which release to pull in for other teams depending on Filecoin-FFI for the coming nv23 upgrade.

PR #483, opened 2024-09-11
chore: update rust version to 1.80.1 and fix lints

Stebalien:


rust/src/proofs/api.rs
```diff
@@ -2068,7 +2068,7 @@ pub mod tests {
             let f_new_sealed_sector = OpenOptions::new()
                 .read(true)
                 .write(true)
-                .create(true)
+                .create_new(true)
```
Stebalien, line 2071: @cryptonemo are these changes correct? Clippy is now complaining about them.

Stebalien:
Going half-way with https://github.com/filecoin-project/filecoin-ffi/pull/483 for now.

Stebalien:
Hm. It looks like I need to fix this even if I only update to 1.78.

Stebalien:
I'm just going to switch to #484.

PR #484, opened 2024-09-13
Update FVM, Deps & Rust

Stebalien:
- Updates the FVM to v4.4, v3.11, and v2.9 in order to pull in the latest wasmtime version and the latest dependencies.
- Update the rest of our dependencies (except those that have breaking changes, I'll do that in a separate PR).
- Update rust to 1.78, the version used/tested in the FVM crate.


rust/src/proofs/api.rs
```diff
@@ -2068,7 +2068,6 @@ pub mod tests {
             let f_new_sealed_sector = OpenOptions::new()
                 .read(true)
                 .write(true)
-                .create(true)
```
Stebalien, line 2071: These files already exist.

Stebalien:
I would ship this in the next feature release, but not in stable as it pulls in new wasmtime versions (risk) and fixes a long-standing bug that only affects macos.

Stebalien:
(although I'd still like to get it merged...)

BigLep:
Message received on the risk here of requiring this during a network upgrade without sufficient bake time in production beforehand.  As a result, we'll include in [v1.31.0](https://github.com/filecoin-project/filecoin-ffi/issues/490).

rvagg (approved):

PR #485, opened 2024-09-13
chore: update safer-ffi

Stebalien:


Stebalien:
On-top-of #484.

cryptonemo:
First thought is to check for memory leaks, as the valgrind command has been removed from CI, but seems reasonable otherwise

Stebalien:
> First thought is to check for memory leaks, as the valgrind command has been removed from CI, but seems reasonable otherwise

Looks like I'm going to need a container for that. Valgrind doesn't work with znver4 optimized distributions.

rvagg (approved):

PR #499, opened 2024-11-07
parallel tests and faster runners

### LINKED ISSUE #498: The "Build and test CGO bindings" unified CI run takes an hour+ on Linux
This is insane.

Stebalien:
fixes #498

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,11 +77,17 @@ jobs:
           go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.59.0
           make go-lint
   cgo-bindings:
-    name: Build and test CGO bindings (${{ matrix.runner }})
-    runs-on: ${{ matrix.runner }}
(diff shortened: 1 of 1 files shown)


Stebalien:
Ok, so... faster. But we don't have a 4x arm64 runner.

BigLep (approved):
Looks good to me?

I assume we're good to no longer have 

`--test-threads 1` and `-p 1`

?

Do we know what changed to allow us to not need this anymore (great that we don't).  (No need to slow down to explain to me - I was just curious).

Stebalien:
> I assume we're good to no longer have
> --test-threads 1 and -p 1

Looking through history, that was added way back in 2019 (possibly before, it could have been ported from rust-fil-proofs-ffi). See https://github.com/filecoin-project/filecoin-ffi/commit/3de4c1567da6cea906a95b87a179c85dbec16c6d. I'm guessing we may have been testing _with_ GPUs at that point and couldn't use the same GPU from multiple threads? I'm not sure, but it doesn't seem necessary now and removing it makes everything faster.

PR #508, opened 2024-11-22
ci: remove unecessary dependencies from macos

Stebalien:
This (pkg-config, specifically) was causing an error in CI due to a conflicting version already being available. It also turns out that `jq` is installed by default in this image as well.


Stebalien:
I assume the upstream builder has changed in some way.

rvagg (approved):

PR #512, opened 2025-01-11
feat: support flush_all_blocks, handle new ipld operations in traces

rvagg:
Ref: https://github.com/filecoin-project/ref-fvm/pull/2101

This is a draft, but the `ExecutionEvent::Log` could probably be done separately since that's much more likely to actually make it in. I'll do that if I end up bailing on this.

### DIFF
--- a/cgo/fvm.go
+++ b/cgo/fvm.go
@@ -90,3 +90,10 @@ func FvmMachineFlush(executor *FvmMachine) ([]byte, error) {
 	}
 	return (SliceBoxedUint8)(resp.value).copy(), nil
 }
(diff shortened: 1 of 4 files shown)


Stebalien (commented):
This seems like a reasonable approach. It's a little annoying that we're losing the ordering of log statements, IPLD events, etc. but that's a limit of the current execution trace format and changing that will be a larger breaking change.

rvagg:
Will leave this one alone until we have a release to use that incorporates it.

Stebalien (approved):

PR #513, opened 2025-01-15
fix: update cache clearing calls

vmx:
The API in `rust-filecoin-proofs-api` was changed, but the public interface of the FFI does not change.


vmx:
Using the released versions of the proofs crates should be enough. This was the breaking change that triggered the 19.0 release of the proofs.

BigLep:
@copilot: make the necessary suggestions to account for the fact that there is a 19.0.0 release of rust-filecoin-proofs-api:  https://github.com/filecoin-project/rust-filecoin-proofs-api/releases/tag/v19.0.0

PR #514, opened 2025-02-02
ci: fix environment configuration on Ubuntu 24.04 arm64

galargh:
This fixes the environment configuration on Ubuntu 24.04 arm64 machines.

The context for the fix is that Ubuntu 24.04 comes with glibc 2.39 installed. Using older versions of nvcc with that glibc version results in the following errors - https://bugs.launchpad.net/ubuntu/+source/mumax3/+bug/2032624. This is specific to arm64 builds only.

On the plus side, we no longer have to downgrade gcc.

As a follow-up piece of work, we should consider installing the latest version of CUDA toolkit in the AMIs.

TODO (@galargh):
- [ ] upgrade 4xlarge self-hosted runners to Ubuntu 24.04

### DIFF
--- a/.github/actions/configure-environment/action.yml
+++ b/.github/actions/configure-environment/action.yml
@@ -4,18 +4,32 @@ description: Configure environment variables for Filecoin FFI
 runs:
   using: 'composite'
   steps:
+    # The installation instructions can be found at https://developer.nvidia.com/cuda-downloads
     - if: runner.os == 'Linux' && runner.arch == 'ARM64'
       run: |
-        wget -q https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2204/sbsa/cuda-ubuntu2204.pin
-        sudo mv cuda-ubuntu2204.pin /etc/apt/preferences.d/cuda-repository-pin-600
-        wget -q https://developer.download.nvidia.com/compute/cuda/12.5.1/local_installers/cuda-repo-ubuntu2204-12-5-local_12.5.1-555.42.06-1_arm64.deb
-        sudo dpkg -i cuda-repo-ubuntu2204-12-5-local_12.5.1-555.42.06-1_arm64.deb
-        sudo cp /var/cuda-repo-ubuntu2204-12-5-local/cuda-*-keyring.gpg /usr/share/keyrings/
+        # Install the CUDA toolkit
+        wget -q https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2404/sbsa/cuda-ubuntu2404.pin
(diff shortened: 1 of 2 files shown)


.github/actions/configure-environment/action.yml
```diff
@@ -4,18 +4,32 @@ description: Configure environment variables for Filecoin FFI
...
       run: |
-        wget -q https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2204/sbsa/cuda-ubuntu2204.pin
-        sudo mv cuda-ubuntu2204.pin /etc/apt/preferences.d/cuda-repository-pin-600
-        wget -q https://developer.download.nvidia.com/compute/cuda/12.5.1/local_installers/cuda-repo-ubuntu2204-12-5-local_12.5.1-555.42.06-1_arm64.deb
-        sudo dpkg -i cuda-repo-ubuntu2204-12-5-local_12.5.1-555.42.06-1_arm64.deb
-        sudo cp /var/cuda-repo-ubuntu2204-12-5-local/cuda-*-keyring.gpg /usr/share/keyrings/
+        # Install the CUDA toolkit
+        wget -q https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2404/sbsa/cuda-ubuntu2404.pin
+        sudo mv cuda-ubuntu2404.pin /etc/apt/preferences.d/cuda-repository-pin-600
+        wget -q https://developer.download.nvidia.com/compute/cuda/12.8.0/local_installers/cuda-repo-ubuntu2404-12-8-local_12.8.0-570.86.10-1_arm64.deb
```
galargh, line 13: I updated the cuda toolkit to the latest version while I was at it.

BigLep:
@rjan90 : are you able to take this one?

rjan90:
@galargh Regarding the missing TODO item about upgrading 4xlarge self-hosted runners to Ubuntu 24.04, do you think this should be addressed before this PR is ready, or handled as a follow-up task?

I have also written up a issue ticket about the follow-up piece of work of installing the latest version of CUDA toolkit in the AMIs: https://github.com/filecoin-project/filecoin-ffi/issues/515

galargh:
> @galargh Regarding the missing TODO item about upgrading 4xlarge self-hosted runners to Ubuntu 24.04, do you think this should be addressed before this PR is ready, or handled as a follow-up task?

@rjan90 I'm ready to go with the upgrade at any moment. I just wanted to first check with you if you're OK with the changes proposed in this PR. If so, I'm going to upgrade the runners and merge this PR.

rjan90:
> 'm ready to go with the upgrade at any moment. I just wanted to first check with you if you're OK with the changes proposed in this PR. If so, I'm going to upgrade the runners and merge this PR.

Yeah, I´m good with it

PR #517, opened 2025-02-18
Fix receivers defined on cgo type aliases

masih:
As of Go 1.24, the compiler will now always fail if receivers are defined on cgo-generated type aliases. Fix the issue by wrapping the cgo-generated types and access to underlying type via `delegate` field. Note that such types cannot be embedded in go structs. Hence, the explicit field.

See:
 * https://tip.golang.org/doc/go1.24#compiler

### DIFF
--- a/cgo/bls.go
+++ b/cgo/bls.go
@@ -9,13 +9,13 @@ package cgo
 import "C"
 
 func Hash(message SliceRefUint8) *[96]byte {
-	resp := C.hash(message)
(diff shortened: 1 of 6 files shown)


Stebalien:
This adds a lot of pointer indirection and that makes me a bit nervous. How about something like https://github.com/filecoin-project/filecoin-ffi/pull/518?

masih:
> How about something like #518?

That works too 👍

PR #522, opened 2025-03-27
chore: release Filecoin-FFI v1.32.0

rjan90:
- Bumps Filecoin-FFI version to v1.32.0
- Updates GST to v0.16.0.
   - I forgot to update to the latest GST release in this PR which shipped the v1.32.0-rc3 FFI release: https://github.com/filecoin-project/filecoin-ffi/pull/520. So just want to sanity check that it should be fine to update this for the final release?


version.json
```diff
@@ -1,3 +1,3 @@
 {
-  "version": "v1.32.0-rc3"
```
masih, line 2: Non-blocker/A small footnote here: I think the way releaser workflow works is that it doesn't include the PR that bumps the version number in the release notes of the release being made. This is because, it creates a draft release _before_ the version bump is merged.

This behaviour is fine when the PR only bumps the version. But _this_ PR does more: it also upgrades dependencies and as a consequence bumps go version.

If you'd like this change to appear in the release note: either separate the PRs, or manually update the generated release note to state this. 

I personally, opt for separate PRs.

masih (approved):

version.json
rjan90: Ah, that is correct. I will separate the PRs

PR #528, opened 2025-04-10
chore: update ref-fvm to v4.7, v3.13, and v2.11

Stebalien:
And update the rest of the dependencies.

### DIFF
--- a/rust/src/fvm/machine.rs
+++ b/rust/src/fvm/machine.rs
@@ -11,7 +11,7 @@ use fvm4_shared::address::Address;
 use fvm4_shared::state::ActorState;
 use fvm4_shared::{ActorID, MethodNum};
 use fvm_ipld_encoding::ipld_block::IpldBlock;
-use fvm_ipld_encoding::tuple::{Deserialize_tuple, Serialize_tuple};
+use fvm_ipld_encoding::tuple::*;
 use fvm_ipld_encoding::{strict_bytes, to_vec, CborStore};
 
 use fvm4_shared::error::{ErrorNumber, ExitCode};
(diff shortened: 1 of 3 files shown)


Stebalien:
Based on #529.

rust/src/fvm/machine.rs
```diff
@@ -11,7 +11,7 @@ use fvm4_shared::address::Address;
 use fvm4_shared::state::ActorState;
 use fvm4_shared::{ActorID, MethodNum};
 use fvm_ipld_encoding::ipld_block::IpldBlock;
-use fvm_ipld_encoding::tuple::{Deserialize_tuple, Serialize_tuple};
+use fvm_ipld_encoding::tuple::*;
```
cryptonemo, line 14: We usually avoid this, but I know you know that and have reasoned about it already

cryptonemo (approved):

rust/src/fvm/machine.rs
Stebalien: Basically, I need to either import the `serde_tuple` re-export or include the crate directly. But if I include the crate directly, I need to make sure that all the versions agree.

The mistake was re-exporting anything from `fvm_ipld_encoding`, but changing that is a larger breaking change.

Stebalien: This became an issue because `serde_tuple` released a v1.0 and we're _not_ updating to that because (a) it doesn't introduce any important new features/fixes and (b) making breaking changes to `fvm_ipld_encoding` is really painful.

Stebalien:
In terms of sizes: this increases the `libfilcrypto.a` size from 122MiB to 133MiB. I think we can get it lower again but we'll need to:

- Update `wasmparser` and friends in `fvm-wasm-instrumentation`.
- Update `rand` (waiting on `k256`).

PR #536, opened 2025-08-13
chore: bump filecoin-proofs-api to v19.0

rjan90:
chore: bump filecoin-proofs-api to v19.0

### Breaking Change Note:

Users will need to update their calls from:
```
// Before
ClearCache(sectorSize, cacheDirPath)
ClearSyntheticProofs(sectorSize, cacheDirPath)

// After  
ClearCache(cacheDirPath)
ClearSyntheticProofs(cacheDirPath)
```
The changes are consistent across all layers and properly remove the unused sector_size parameter completely from the API.

### DIFF
--- a/cgo/proofs.go
+++ b/cgo/proofs.go
@@ -322,17 +322,15 @@ func GetNumPartitionForFallbackPost(registeredProof RegisteredPoStProof, numSect
 	return uint(resp.value), nil
 }
 
-func ClearCache(sectorSize uint64, cacheDirPath SliceRefUint8) error {
+func ClearCache(cacheDirPath SliceRefUint8) error {
 	resp := (*resultVoid)(C.clear_cache(
-		C.uint64_t(sectorSize),
 		(C.slice_ref_uint8_t)(cacheDirPath)))
 	defer resp.destroy()
 	return CheckErr(resp)
 }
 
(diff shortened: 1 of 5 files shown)


rvagg:
looks like we removed `sector_size` from a couple of calls that'll need to be updated here

rjan90:
> oh .. those sectorSize arguments go all the way up to the Go API, we should remove it there as well and this would become a breaking change.

Ah, just pushed underscoring the parameters, as I believed we still would need them for the Go-code. But I will look into removing them there

vmx:
See  https://github.com/filecoin-project/filecoin-ffi/pull/513 for how I've planned it. Though a breaking change is even better if that's possible.

rjan90:
@vmx I was also going down that route in the first couple of commits:
- https://github.com/filecoin-project/filecoin-ffi/pull/536/commits/1a66f3c49c8705972fe3284752e6772bcd87ffa8
- https://github.com/filecoin-project/filecoin-ffi/pull/536/commits/a823243b7c18a60f7469a0818874b9b7c0e5b59c

But the build checks were still failing after those changes. At which point I switched to remove the sectorSize from `ClearCache` and `ClearSyntheticProof`.

rvagg (approved):

PR #542, opened 2025-09-16
ci: create release uploader workflow

galargh:
This will create a new workflow `Upload Release Assets` which can be used to do emergency release assets uploads when something goes wrong.

I tested it by triggering https://github.com/filecoin-project/filecoin-ffi/actions/runs/17773290672 on push. It is failing because the action I'm adding here is not available in the tag of the release that we were uploading assets to. It's just a false negative.

### DIFF
--- a/.github/actions/upload-release-assets/action.yml
+++ b/.github/actions/upload-release-assets/action.yml
@@ -1,0 +1,73 @@
+name: Configure Environment Variables
+description: Configure environment variables for Filecoin FFI
+
+inputs:
+  id:
+    description: The release ID
+    required: true
+  ref:
+    description: The release ref
+    required: true
+
+runs:
+  using: 'composite'
+  steps:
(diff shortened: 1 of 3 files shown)


BigLep:
@galargh : I updated RELEASE.md: https://github.com/filecoin-project/filecoin-ffi/pull/542/commits/cae19d6623cbbcd260fef41ae9c568f1a469e5e9

It seems like this improvement idea would have been useful here: "Add a check to the [Releaser](.github/workflows/release.yml) workflow to ensure that the created/published release contains the expected assets. If it does not, create them and run the [publish-release.sh](rust/scripts/publish-release.sh) script to upload the missing assets."

BigLep:
> Yes, I like it! Let's do it as a separate item though as it will require adding a check for whether the release assets are already there which is new.

Issue created (https://github.com/filecoin-project/filecoin-ffi/issues/543 ) and I assigned it to @copilot to see how it does...  Feel free to take over.

I'll squash and merge.

PR #545, opened 2025-09-21
chore!: upgrade to Go 1.24

galargh:
## Proposed Changes

This PR updates Go version used in this repo to 1.24.7.

## Additional Info

golangci-lint v1.64 is the last version in the v1 lineage. It has moved on to v2 now. v1.64 does not support Go 1.25 so we'll have to upgrade to v2 before we move on to Go 1.25 ourselves.

### DIFF
--- a/.github/actions/configure-environment/action.yml
+++ b/.github/actions/configure-environment/action.yml
@@ -63,4 +63,4 @@ runs:
         toolchain: 1.73
     - uses: actions/setup-go@v5
       with:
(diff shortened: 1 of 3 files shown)


galargh:
Right on, I'll get on it tomorrow!

galargh:
@rvagg Could you point me in the direction of which types are generated here/how to regenerate them if needed? `make` still works both in the CI and on my local machine.

PR #586, opened 2026-09-07
fix(deps): use curiostorage/sppark fork

snadrus:
The fix for auto-detect breaking on SM80 is not getting picked-up by Supraseal. It appears like a dead repo, so the repo was forked and the change added. 
This enables older hardware to be auto-detected correctly.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -3344,8 +3344,7 @@ dependencies = [
 [[package]]
 name = "sppark"
 version = "0.1.15"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


LexLuthr (approved):
Should we ship the crate or keep it in guthub?

LexLuthr:
@snadrus Can we also explore using forked SupraSeal for FFI? We have had similar issues with upgrades to that as well.

snadrus:
> @snadrus Can we also explore using forked SupraSeal for FFI? We have had similar issues with upgrades to that as well.

I figure once we have the next significant change that we will need to do that.

Issue #587, opened 2026-09-16
Release with ref-fvm 4.8.3 (nv29-dev flag removed) for nv29 Solstice


BigLep:
## Done Criteria

- [x] filecoin-ffi `rust/Cargo.toml` depends on `fvm` / `fvm_shared` `~4.8.3` with the `nv29-dev` feature removed, and `rust/Cargo.lock` is updated to match. (#588; unblocked by filecoin-project/ref-fvm#2345, which let fvm 4.8.3 be published to crates.io)
- [x] A filecoin-ffi release is published containing that change (version string per discussion below). (#589; [v1.37.0 release](https://github.com/filecoin-project/filecoin-ffi/releases/tag/v1.37.0), published 2026-09-16, tag commit f9b4aec)
- [x] Lotus `master` has `go.mod` and the `extern/filecoin-ffi` submodule pointing at that release, per [Update_Dependencies_Lotus.md](https://github.com/filecoin-project/lotus/blob/master/documentation/misc/Update_Dependencies_Lotus.md#updating-filecoin-ffi). (filecoin-project/lotus#13804, merged 2026-09-17; submodule at the v1.37.0 tag commit f9b4aec)
- [x] This issue is linked from the Lotus v1.37.0 release issue filecoin-project/lotus#13760. (already listed in that issue's dependency checklist)

## Why Important

This is the last dependency for the nv29 (Solstice) upgrade that is still on a `-dev` gated build. Lotus `master` currently ships filecoin-ffi v1.36.1, which pins `fvm ~4.8.2` with `features = ["nv29-dev"]`. ref-fvm 4.8.3 (released 2026-09-14) removed the `nv29-dev` feature flag in filecoin-project/ref-fvm#2336, so nv29 support is now unconditional in a stable FVM release. Shipping the final nv29 Lotus on an FVM that only supports nv29 behind a dev flag is not what we want for mainnet, and once the flag is gone FFI cannot pick up 4.8.3 by lockfile bump alone: requesting a feature that no longer exists fails to resolve. Hence the one-line `Cargo.toml` change plus a release.

## User/Customer

Implementers (Lotus, and anyone else consuming filecoin-ffi for nv29).

## Why this is safe

The `nv29-dev` flag gated exactly two things in ref-fvm, both visible in the [4.8.2 to 4.8.3 diff](https://github.com/filecoin-project/ref-fvm/compare/fvm@v4.8.2...fvm@v4.8.3):

[...]

PR #588, opened 2026-09-16
chore(deps): bump fvm to ~4.8.3 and drop nv29-dev feature

BigLep:
Closes the first done criterion of #587.

## What

- `rust/Cargo.toml`: `fvm` / `fvm_shared` from `~4.8.2` to `~4.8.3`, and remove the `nv29-dev` feature from `fvm4`.
- `rust/Cargo.lock`: `cargo update -p fvm -p fvm_shared`. Besides the two fvm crates, this moves `cid` 0.11.1 to 0.11.3 and `multihash` 0.19.3 to 0.19.5 (with its derive crates), which are the minimums fvm 4.8.3 requires, and drops the now-unused `core2`.

ref-fvm 4.8.3 removed the `nv29-dev` flag in filecoin-project/ref-fvm#2336, so nv29 (Solstice) support is unconditional. Requesting a feature that no longer exists fails to resolve, so the feature must be dropped together with the version bump. See #587 for why this is safe.

## Verification

`cargo check --no-default-features --features fvm` completes cleanly on macOS arm64 against the published crates. No filecoin-ffi source changes are needed for 4.8.3. CI covers the full default-feature build.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


BigLep:
After this lands, I plan to do a 1.37.0 release following https://github.com/filecoin-project/filecoin-ffi/blob/master/RELEASE.md

rvagg:
I've also added in the v2 and v3 crates here as well since they were also released: fvm@v2.12.2/fvm_shared@v2.12.2 & fvm@v3.14.2/fvm_shared@v3.14.2

rvagg (approved):

